PR #3, opened 2020-05-04
Upgrade zeroconf to 0.26.0

fabaff:
To be in sync with Home Assistant -> https://github.com/home-assistant/core/pull/35176

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ classifiers = [
 [tool.poetry.dependencies]
 python = "^3.7"
 cryptography = "^2.8"
-zeroconf = {version = "^0.24.4", extras = ["IP"]}
+zeroconf = {version = "^0.26.0", extras = ["IP"]}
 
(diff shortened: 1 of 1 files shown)


Jc2k:
Cheers @fabaff. Is this tripping something up in HA? Wondering if there is a better version specifier I should be using than `^`. Looking at the [docs](https://python-poetry.org/docs/versions/) I had assumed`^0.24.4` should be like `>=0.24.4, <1` but reading a bit more it looks like its actually `>=0.24.4,<0.25`, which is a bit annoying 😭 

I'd be happy to change the specifier to `>=0.26.0` and then hopefully this won't require a new release every time HA wants to bump zeroconf. What do you think? The risk for homekit_controller is that zeroconf break their ABI in a future release, but I don't want to be on the critical path so it's probably a risk i'm happy to take.

If I did that i'd probably still want to bump my versions to match HA periodically, but it wouldn't be a problem for HA if i was behind.

PR #6, opened 2020-07-09
Allow using shared zeroconf

bdraco:


Jc2k:
Whoop. This areas a bit of a mess (hold overs from its roots in the previous HK library) - so i'm sorry you've had to fight with it, but super excited by the results. When the tests pass i'm happy to merge this.

My hope is that we can also actually stop causing quite so many discoveries from within aiohomekit.

The most obvious candidate for further improvement is when HA passes a device to aiohomekit that it discovered itself to aiohomekit. Right now it passes a device id and aiohomekit does its own discovery. The fact that aiohomekit needs to do its own discovery is just a throwback to the previous library that aiohomekit is based on. I didn't want to try and get an API change upstream. But HA is the only customer of this library so I really want to fix that and add some API that lets HA just pass what it found over directly without any additional discovery steps.

The other is device a discovery before connecting to a device. Right now thats needed because some devices randomly change port during the day. But we should be able to passively listen to the announcements and update the config flow when we see a port change. And maybe do a manual discovery if the first connection attempt fails.

bdraco:
Having trouble getting `isort` to agree.  Probably different versions.

bdraco:
@Jc2k I think its good to go, I'm not sure why the deploy step is failing or if it should be running for PRs??

Jc2k:
It definitely shouldn't be. Sigh. I'll look at that but for now i'm happy this is ready.

PR #7, opened 2020-07-09
Controller takes a zeroconf_instance

Jc2k:
@bdraco I thought about it some more and we didn't take into account [this](https://github.com/Jc2k/aiohomekit/blob/master/aiohomekit/controller/ip/connection.py#L530) case.

That's one of the ones that i'd like to be able to get rid of in the long run (or at least not call as often), but right now it's needed for some devices to stay working for more than a day.

The most natural way to get to the zeroconf_instance from that part of the code would be if there was a zeroconf_instance on controller i think, so thats what this PR does.

What do you think?

### DIFF
--- a/aiohomekit/controller/controller.py
+++ b/aiohomekit/controller/controller.py
@@ -40,17 +40,18 @@ class Controller(object):
     This class represents a HomeKit controller (normally your iPhone or iPad).
     """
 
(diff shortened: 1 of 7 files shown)


bdraco:
I think this is better approach since you'll always have the zeroconf instance in advance 👍

PR #9, opened 2020-07-30
Fixes

ntninja:
While trying to get home-assistant to talk to [hap](https://lib.rs/crates/hap), I've so far uncovered two grave issues with this library that I've fixed:

  * aiohomekit's homebrew HTTP parser fails to notice any non-titlecase HTTP headers (this is a spec violation – HTTP headers are case-insensitive)
  * When a connection attempt times out, an asyncio `_reconnect` task will be left spinning in the background

I may update this if I find more issues to be fixed.

### DIFF
--- a/aiohomekit/controller/ip/connection.py
+++ b/aiohomekit/controller/ip/connection.py
@@ -246,7 +246,10 @@ class HomeKitConnection:
         if not self._connector:
             return
         self._connector.cancel()
-        await self._connector
+        try:
+            await self._connector
+        except asyncio.CancelledError:
+            pass
         self._connector = None
 
     async def get(self, target):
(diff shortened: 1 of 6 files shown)


Jc2k:
This is brilliant. Thank you! As it seems you aren't done testing I'll hold off merging. Let me know when you are done?

ntninja:
Thanks! I still need to figure out a test for this, but in the meantime could you take a look at the macOS build error? It does not look like anything I could directly influence.

Jc2k:
It's probably just a flaky test - will rerun in morning.

ntninja:
@Jc2k: Done! (macOS still fails though.)

PR #19, opened 2020-08-24
Add license text

fabaff:
Shipping the license is a requirement for distribution packages. This is something that would come up during th Fedora review process.

It would be nice if you could publish a new release after merging this? Thanks

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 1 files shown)


Jc2k:
Cheers @fabaff. Happy to do anything to support distro packaging, but be warned aiohomekit is a rapidly moving target! :-)

Jc2k:
0.2.50 tagged. As long as CI doesn't flake there should be a wheel in a few minutes.

fabaff:
Thanks for the quick action.

Review request (as reference): https://bugzilla.redhat.com/show_bug.cgi?id=1871908

PR #23, opened 2020-10-24
add vocolinc flowerbud chars

adrum:
[This product](https://www.vocolinc.com/products/flowerbud-smart-diffuser) has some custom characteristics to operate as a diffuser, despite being classified as a humidifier. This adds the ability to use these characteristics.

### DIFF
--- a/aiohomekit/model/services/service.py
+++ b/aiohomekit/model/services/service.py
@@ -41,7 +41,10 @@ class Characteristics:
         matches = iter(self)
 
         if char_types:
-            char_types = [(c if len(c) == 36 else CharacteristicsTypes.get_uuid(c)) for c in char_types]
+            char_types = [
+                (c if len(c) == 36 else CharacteristicsTypes.get_uuid(c))
+                for c in char_types
+            ]
             matches = filter(lambda char: char.type in char_types, matches)
 
         return matches
--- a/aiohomekit/model/services/service.py
+++ b/aiohomekit/model/services/service.py
@@ -41,10 +41,7 @@ class Characteristics:
         matches = iter(self)
 
         if char_types:
-            char_types = [
-                (c if len(c) == 36 else CharacteristicsTypes.get_uuid(c))
-                for c in char_types
-            ]
+            char_types = [CharacteristicsTypes.get_uuid(c) for c in char_types]
             matches = filter(lambda char: char.type in char_types, matches)
 
         return matches
--- a/aiohomekit/model/services/service.py
+++ b/aiohomekit/model/services/service.py
@@ -41,7 +41,7 @@ class Characteristics:
         matches = iter(self)
 
         if char_types:
-            char_types = [CharacteristicsTypes.get_uuid(c) for c in char_types]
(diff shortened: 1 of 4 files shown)


aiohomekit/model/services/service.py
```diff
@@ -41,7 +41,10 @@ def filter(self, char_types=None) -> Iterable[Characteristic]:
         matches = iter(self)
 
         if char_types:
-            char_types = [CharacteristicsTypes.get_uuid(c) for c in char_types]
+            char_types = [
+                (c if len(c) == 36 else CharacteristicsTypes.get_uuid(c))
```
Jc2k, line 45: Can we do this normalisation in get_uuid?

adrum: Yes, I originally had it there. I will add it back and update the tests.

Jc2k:
Are you planning to add these to HA? There is not really a good story for vendor extensions there yet. I guess the vague plans are:

* Vendor extensions that can be modelled as standalone sensors and binary_sensors (and then related to the main entity via the device registry) should be. (There was a recent decision that even the "battery charging" attribute should be an entity in its own right. Because of that decision I won't be adding any more attributes without input from HA core devs). This is complicated because right now. There is an assumption of 1 entity per service. So some refactoring will be required. I have a branch in progress for this but it was more work than I had time for at the time.

* Potentially allow some of the more complicated entity types like climate to be subclassed, loading the subclass if all of its required vendor extensions are detected. This would allow some thermostats the ability to use presets, for example. I do not want to have vendor specific logic in the main entities.  Again, refactoring will be needed here.

* While I am fine with using vendor extensions to implement existing HA interfaces/patterns/services/etc I am really really reluctant to add support for vendor extensions that don't map to anything on the HA side and can only be interacted via vendor specific service calls, especially especially for devices I don't have.

aiohomekit/model/services/service.py
Jc2k: Brilliant, thank you. When that's done I can push a tag/pypi release with it merged for you if you'd like?

adrum:
@Jc2k I am planning on utilizing this in HA. I have 2 PRs pretty much ready to go, but is related to this PR and utilizes vendor specific chars.

- The first one is to support a per HAP spec Humidifier/Dehumidifier. 

- The other will be to support this specific VOCOLinc product, as an alternative Humdifier/Dehumidifer. It won't have any vendor specific service calls associated with it, and will work only by utilizing the HA humidifier controls. The PR only leverages one of the chars introduced here, but I went ahead and threw in the rest of the chars.

They both will map to the humdifier entity in HA.

Jc2k:
Nice! Sounds like it aligns with my thinking on case 2 then! Great news, i look forward to reviewing those PR's as well!

Jc2k (approved):

Jc2k:
I've run out of time tonight but I'll try and push out a tag tomorrow.

adrum:
@Jc2k Can you tag a new release? I have the PR ready for this in HA.

PR #30, opened 2021-05-12
Load from the zeroconf cache when a ServiceBrowser for _hap is running

bdraco:
Fixes https://github.com/home-assistant/core/issues/50534
Requires https://github.com/jstasiak/python-zeroconf/pull/356

Since there is a ServiceBrowser already running collecting the records in HomeAssistant everything is now near instant similar to the experience the user would get in the Home app:

5YOU8qszHl

### DIFF
--- a/aiohomekit/zeroconf/__init__.py
+++ b/aiohomekit/zeroconf/__init__.py
@@ -22,7 +22,7 @@ from functools import partial
 import logging
 import threading
 from time import sleep
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 from zeroconf import Error, ServiceBrowser, ServiceInfo, Zeroconf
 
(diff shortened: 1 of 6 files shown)


Jc2k:
Thanks for this @bdraco! Not familiar enough with zeroconf internals and the work you have been doing in HA to review it as thoroughly as I would normally want to, but am more than happy to defer to you on this and take whatever you think is needed :+1:

bdraco:
Just realized there is another one in aiohomekit.controller.up.zeroconf    

I'll adjust that one as well and move the function (and import it back to the existing location for compat) into aiohomekit.zeroconf so it's easier to avoid missing in the future.

aiohomekit/zeroconf/__init__.py
```diff
@@ -91,6 +91,24 @@ def get_data(self) -> List[ServiceInfo]:
...
+) -> Dict[str, Any]:
+    """Find a homekit device in the zeroconf cache."""
+    for name in zeroconf.cache.names():
+        if not name.endswith(HAP_TYPE):
+            continue
+        info = ServiceInfo(HAP_TYPE, name)
+        info.load_from_cache(zeroconf)
+        if b"id" not in info.properties:
+            continue
+        if info.properties[b"id"].decode() == device_id:
```
bdraco, line 105: Would be slightly more efficient to make the device id bytes so we don't have to decode each loop

Jc2k:
> Just realized there is another one in aiohomekit.controller.up.zeroconf    
> 
> I'll adjust that one as well and move the function (and import it back to the existing location for compat) into aiohomekit.zeroconf so it's easier to avoid missing in the future.

Sounds good. That code is a bit messed up from when I forked homekit_python, never finished cleaning up after Asyncio-ification (was originally meant to be a PR to add async to homekit_python). So feel free to restructure things as you see fit, I'm not wedded to it's current form at all.

bdraco:
I moved `async_discover_homekit_devices` but it didn't have coverage so I'll need to add tests for that

bdraco:
Once I finished up https://github.com/home-assistant/core/pull/50784, I'll come back to this

bdraco:
@Jc2k This should be ready to go now that https://github.com/home-assistant/core/pull/50807 is bumped

bdraco:
@Jc2k *gentle nudge* Any chance you can take a look so we can get this in before beta?

Jc2k:
I'm not going to have time to do any testing by then myself but I trust you/your work enough to merge and release and leave the testing entirely to you, if that works for you?

bdraco:
> I'm not going to have time to do any testing by then myself but I trust you/your work enough to merge and release and leave the testing entirely to you, if that works for you?

Sounds good.  I've tested this on multiple/systems network quite a bit and have been using it in production for for the greater part of a week.

Jc2k:
@bdraco there you go - new tag on pypi

bdraco:
Thanks 👍

PR #31, opened 2021-05-22
Subscribe characteristics one at a time to match iOS behavior

bdraco:
Fixes https://github.com/home-assistant/core/issues/37996

### DIFF
--- a/tests/accessoryserver.py
+++ b/tests/accessoryserver.py
@@ -829,19 +829,17 @@ class AccessoryRequestHandler(BaseHTTPRequestHandler):
         if changed:
             self.server.write_event(changed, self.session_id)
 
-        if len(result["characteristics"]) == errors:
-            self.send_response(HttpStatusCodes.BAD_REQUEST)
-        elif len(result["characteristics"]) > errors > 0:
-            self.send_response(HttpStatusCodes.MULTI_STATUS)
-        else:
+        if errors == 0:
             self.send_response(HttpStatusCodes.NO_CONTENT)
             self.end_headers()
-        if errors > 0:
-            result_bytes = json.dumps(result).encode()
-            self.send_header("Content-Type", "application/hap+json")
-            self.send_header("Content-Length", len(result_bytes))
-            self.end_headers()
-            self.wfile.write(result_bytes)
+            return
+
+        result_bytes = json.dumps(result).encode()
+        self.send_response(HttpStatusCodes.MULTI_STATUS)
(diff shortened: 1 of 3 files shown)


tests/accessoryserver.py
```diff
@@ -829,19 +829,17 @@ def _put_characteristics(self):
         if changed:
             self.server.write_event(changed, self.session_id)
 
-        if len(result["characteristics"]) == errors:
-            self.send_response(HttpStatusCodes.BAD_REQUEST)
```
bdraco, line 833: The spec says we MUST response with 207 if there is any error

bdraco:
Manually verified subscriptions still work with a VOCOlinc-SmartBar

bdraco:
Tested with MyQ bridge as well

bdraco:
I'm not sure if this is needed for unsubscribe as well but it seems like if the device has trouble with subscribe it would have the same issue with unsubscribe.   


Not sure if that actually matters ?

bdraco:
Looks like we do await self.pairing.unsubscribe(self.watchable_characteristics) in async_unload. 

I'm not sure we need to do that when we unload as closing the connection should invalidate all the event subscriptions?   I can't actually find where we close the connection though.

Jc2k:
I think you are probably right. I added the unsubscribe to be symmetrical, rather than because i saw iOS do it. I guess its worth checking if iOS does just close the connection? If it does then i'm sold, otherwise we should do what it does.

And you are right, it is missing an `await pairing.close` in `async_unload`. I think I tried to clean up after stuff that I added but the original homekit_controller never closed it's connection, and I never noticed before. Ugh!

Jc2k (approved):

Jc2k:
I'm going to be entirely AFK today I think, so I'm going to merge and release this so you aren't blocked. If iOS does unsubscribe before it closes the connection I guess we'll need another PR, but i don't mind doing extra releases.

Thanks for sorting this out as well!

Jc2k:
New tag out on pypi.

bdraco:
iOS closes the connection and does not unsubscribe in all cases I could test. 

In fact it does everything it can (prevents underlying RST from happening) to not even bother to close the connection in the hopes it can reuse it hours later. 

This behavior is why we have a cleanup check at 90 hours in HAP-python https://github.com/ikalchev/HAP-python/blob/dev/pyhap/hap_protocol.py#L18

bdraco:
> New tag out on pypi.

Thanks!

PR #32, opened 2021-05-23
Make close an async def for FakePairing testing

bdraco:


Jc2k (approved):

bdraco:
I think the windows test here is flakey.

I'll amend and repush to make sure its just a transient failure

Jc2k:
Yes there is a flakey test. It flakes most on Windows, but I have seen it flake on them all 😭

PR #35, opened 2021-05-25
Subscribe/Unsubscribe one accessory at a time

bdraco:
https://github.com/Jc2k/aiohomekit/pull/31 changed the behavior to subscribe to one char at a time for subscribe.  That's really inefficient, but closer to iOS.

This PR takes a more efficient approach which is to subscribe/unsubscribe by doing the one call per accessory id at a time which matches iOS 14.6 behavior.

### DIFF
--- a/aiohomekit/controller/ip/pairing.py
+++ b/aiohomekit/controller/ip/pairing.py
@@ -15,8 +15,10 @@
 #
 
 import asyncio
+from itertools import groupby
 import json
 import logging
+from operator import itemgetter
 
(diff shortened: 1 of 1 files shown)


bdraco:
Tested with a smart bar and a myq bridge 👍

bdraco:
Tested with a heated floor TH1310WF thermostat 👍

bdraco:
Tested with Tado Smart AC Control 👍

Jc2k (approved):
Nice 👍

Issue #10, opened 2020-07-30
[Suggestion] Use h11 Python library for HTTP/1.1 header parsing


ntninja:
One of the issues fixed in #9 could have easily be prevented by not employing a roll-your-own HTTP header parsing library. While the custom framing requirements making using a full-blown HTTP client undesirable, at the HTTP parsing/generation could be delegated to a common library maintained by “somebody else” and [h11](https://pypi.org/project/h11/) seems like the perfect tool for this job as it (I know at least one “serious” HTTP library that uses it).

Jc2k:
Interesting. Will it handle a stream of interspersed EVENT/1.0 and HTTP/1.1 messages? Regarding the generation part it really depends how much control it gives, will have a look soon. We have to talk to a lot of very broken hap implementations, and often that means trying to be byte for byte compatible with what iOS does. Aren't there yet with that but it's looking more and more inevitable.

Not applicable to HTTP handling but e.g. recently had a case where a thermostat couldn't handle a space in a valid JSON body. (Actually only some API endpoints were affected by this). A different thermostat would accept 1, 0 and false for a boolean field. But not true. The spec says all 4 are valid, but we now have to send 1 or 0 only which is what the implementation does. We've had some issues with the Host header in the past too.

So try to be as close to the iOS implementation as possible for sending and as forgiving as possible with reading.

There is a lot of stuff like it is because this is a fork of homekit_python. The plan was to do quite a bit of refactoring but my time has been swallowed up so I'm not really actively improving it as I had hoped just reacting to bugs. So code improvements very much welcome.

Jc2k:
As suspected, it looks h11 would not work for this particular HTTP dialect (directly). E.g.

https://github.com/python-hyper/h11/blob/master/h11/_abnf.py#L84

This is used in a regex here (via `status_line`, defined a few lines below `http_version`):

https://github.com/python-hyper/h11/blob/41a3467dd98271529cce361c756da9e34d8e2b00/h11/_readers.py#L89

As you probably know, when a HomeKit accessory notifies a controller about an event it violates the HTTP spec and sends a response without a request. This is one of the other reasons we don't use an existing client - they don't tend to like that as they expect strict request/response. As you say it would be nice to use the parsing. But it's not a `HTTP/1.1` message, its an `EVENT/1.0` message. In every other way, it is `HTTP/1/1` compatible, apart from most parsers don't expect the 5 bytes to be anything other than `HTTP/`.

[...]

PR #37, opened 2021-05-31
Better handling of newlines inside http response body

Jc2k:
As https://github.com/Jc2k/aiohomekit/issues/10 is still not viable, this continues to adapt our internal parsing.

This may be a fix for https://github.com/home-assistant/core/issues/46785.

This is a pretty core part of aiohomekit, changes here could subtly break things for many real world devices. So i've treated lightly. I changed the chunked encoding handling as little as possible while pulling it out of the header parsing code.

## The bug

If there is a new line in the body it hits the bit of code inside the while loop.

```
                if self._content_length > -1:
                    self.body += self._raw_response
                    self._raw_response = bytearray()
```

The blindly appends ALL the remaining buffer, even if content length was 0. Crucially it might take data it doesn't own, corrupting the parser state. The buffer is now empty and it doesn't hit the second code path.

If there isn't a newline it hits the 2nd code path only:

```
            remaining = self._content_length - len(self.body)
            self.body += self._raw_response[:remaining]
            self._raw_response = self._raw_response[remaining:]
```

This tries to drain only enough from the buffer to satisfy the remainder. So this code path works. This creates the erratic behaviour you see in the camera previews.

In addition, the while loop is destructive. Going in there consumes a "line". So we can't go in there if self._state == STATE_BODY. So pull chunked encoding out of that loop.

### DIFF
--- a/aiohomekit/controller/ip/pairing.py
+++ b/aiohomekit/controller/ip/pairing.py
@@ -229,7 +229,7 @@ class IpPairing(AbstractPairing):
             await self.list_accessories_and_characteristics()
 
         url = "/characteristics?id=" + ",".join(
-            [str(x[0]) + "." + str(x[1]) for x in characteristics]
+            str(x[0]) + "." + str(x[1]) for x in characteristics
         )
         if include_meta:
             url += "&meta=1"
--- a/aiohomekit/http/response.py
+++ b/aiohomekit/http/response.py
@@ -45,7 +45,7 @@ class HttpResponse:
         self._raw_response += part
         pos = self._raw_response.find(b"\r\n")
 
-        while pos != -1 and self._state < HttpResponse.STATE_DONE:
+        while pos != -1 and self._state < HttpResponse.STATE_BODY:
             line = self._raw_response[:pos]
             self._raw_response = self._raw_response[pos + 2 :]
             if self._state == HttpResponse.STATE_PRE_STATUS:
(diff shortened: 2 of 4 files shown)


bdraco:
I tried the new code and verified its fixed.

Does it make sense to copy/reuse the reader/buffer code from `h11` (has a compatible license as long as MIT license is included/authored given credit) since they have likely had to solve all the corner cases?

https://github.com/python-hyper/h11/blob/master/h11/_receivebuffer.py
https://github.com/python-hyper/h11/blob/master/h11/_readers.py

Jc2k:
It's horrible code (in aiohomekit) and i'd love to replace it. But:

* I don't really want to make aiohomekit mixed licence.
* I'm confident tha they have made a formally correct implementation of HTTP, i'm not confident all these device vendors have. We certainly know that the device vendors can't even handle valid JSON in some cases. And there are edge cases in h11 where its behaviour is known to be stricter than browsers and curl. Some of these edge cases we may well have to "fix". So I'd want to collect quite a lot more HTTP parsing tests cases from real devices before doing it given the risk of regressions.
* Although the code upsets me, it's not been that big a source of bugs. I feel like maybe 3 or 4 in a couple of years? And the camera stuff is a pretty recent endpoint that support was added for. So i'm reluctant to replace such a vital low level bit given a risk of regressions. This is also why i've been reluctant to refactor this bit, even though it could do with it. Would like more tests either way, I guess!

Jc2k:
Will merge and release this as soon as I get chance.

bdraco:
> It's horrible code (in aiohomekit) and i'd love to replace it. But:
> 
> * I don't really want to make aiohomekit mixed licence.
> * I'm confident tha they have made a formally correct implementation of HTTP, i'm not confident all these device vendors have. We certainly know that the device vendors can't even handle valid JSON in some cases. And there are edge cases in h11 where its behaviour is known to be stricter than browsers and curl. Some of these edge cases we may well have to "fix". So I'd want to collect quite a lot more HTTP parsing tests cases from real devices before doing it given the risk of regressions.
> * Although the code upsets me, it's not been that big a source of bugs. I feel like maybe 3 or 4 in a couple of years? And the camera stuff is a pretty recent endpoint that support was added for. So i'm reluctant to replace such a vital low level bit given a risk of regressions. This is also why i've been reluctant to refactor this bit, even though it could do with it. Would like more tests either way, I guess!

Makes sense.  The almost http/1.1 but not... HAP `feature` seems like a bit of a miss on apple's part since so many devices end up getting it wrong.  I wonder if having a side channel for events would have saved lots of device devs a lot of headaches.

PR #38, opened 2021-06-03
Disable using the ServiceBrowser cache

bdraco:
zeroconf 0.32.0 (unreleased) is needed to ensure compatibility with some devices


bdraco:
I don't think we are going to need this one as it looks like the root cause the was `C#` issue.

https://github.com/Jc2k/aiohomekit/pull/39 will take care of the sometimes missing `addresses` issue but I'm hoping we don't see any reports of that in Home Assistant since the problem has existed for some time but is likely made a bit worse due to relying more heavily on the cache.

PR #39, opened 2021-06-03
Update to use AsyncZeroconf

bdraco:
Needs zeroconf 0.32.0 which should hopefully be released soon.

With zeroconf 0.32.0, a new `AsyncServiceBrowser` has been introduced.  This PR teaches aiohomekit about it so it does not fallback to slow discovery because it only knew about `ServiceBrowser`

This should fix all the inconsistencies with devices that don't provide the A records in the additionals fields when requesting PTRs

Fixes https://github.com/home-assistant/core/issues/51851

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ classifiers = [
 python = "^3.7"
 cryptography = ">=2.9.2"
 zeroconf = ">=0.32.0"
+asynctest = "^0.13.0"
 
 [tool.poetry.dev-dependencies]
 isort = "^5.4.2"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,6 @@ classifiers = [
 python = "^3.7"
 cryptography = ">=2.9.2"
 zeroconf = ">=0.32.0"
-asynctest = "^0.13.0"
 
 [tool.poetry.dev-dependencies]
 isort = "^5.4.2"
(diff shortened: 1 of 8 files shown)


bdraco:
Needs https://github.com/jstasiak/python-zeroconf/issues/643

bdraco:
@Jc2k  This should resolve the performance regression with zeroconf 0.32.0 since homekit_controller will now know about `AsyncServiceBrowser`.  Hopefully this looks good and can be released quickly so we can get it into beta.

pyproject.toml
```diff
@@ -20,7 +20,8 @@ classifiers = [
 [tool.poetry.dependencies]
 python = "^3.7"
 cryptography = ">=2.9.2"
-zeroconf = ">=0.31.0"
+zeroconf = ">=0.32.0"
+asynctest = "^0.13.0"
```
Jc2k, line 24: I'm guessing `asynctest` should be in the dev-dependencies section? (`poetry add -D asynctest`).

Jc2k (approved):
Had a quick scan and seemed to make sense to me, but as ever with the zeroconf stuff i'm going to defer to your expertise (+ your testing setup is so much more extensive than mine).

Happy to put out a 0.4.0 with this in if you think we are ready?

pyproject.toml
bdraco: Finished testing on 3 production installs.  

I'll push a commit to move it in a second

bdraco: @Jc2k Should be good to go for 0.4.0 👍

PR #42, opened 2021-07-03
Exclude _hap._tcp.local. when enumerating the zeroconf cache

bdraco:
When a ServiceBrowser is running, we check the cache for existing
_hap._tcp.local. entries. The cache enumeration failed to exclude
_hap._tcp.local. itself (service type with no name) which meant
extra unneeded questions for _hap._tcp.local. itself would be asked
along with each NAME._hap._tcp.local. that needed to be refreshed.

### DIFF
--- a/aiohomekit/zeroconf/__init__.py
+++ b/aiohomekit/zeroconf/__init__.py
@@ -85,7 +85,7 @@ async def _async_homekit_devices_from_cache(
     """Return all homekit devices in the cache, updating any missing data as needed."""
     infos = []
     for name in aiozc.zeroconf.cache.names():
-        if not name.endswith(HAP_TYPE):
+        if not name.endswith(HAP_TYPE) or name == HAP_TYPE:
             continue
         infos.append(AsyncServiceInfo(HAP_TYPE, name))
 
--- a/tests/test_zeroconf.py
+++ b/tests/test_zeroconf.py
@@ -16,7 +16,7 @@
 
 import socket
 import sys
-from unittest.mock import MagicMock, PropertyMock, patch
+from unittest.mock import MagicMock, PropertyMock, call, patch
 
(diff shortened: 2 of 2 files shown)


bdraco:
@Jc2k I was doing validation and watching all the zeroconf traffic line by line in beta and noticed this.  It doesn't actually cause any noticeable problem, but it does generate unneeded mdns traffic.

Sorry I missed this one the first round.

Jc2k (approved):
LGTM

Jc2k:
Will put out a .1 in the morning (couple of hours)

bdraco:
@Jc2k I found another issue in testing.  Tado devices don't use case-intensive dns names so they never respond if to the cached named if its been lowercased.  I'm going to see if I can fix that as well

Jc2k:
Ah good spot. Will hold off the release then. Cheers.

bdraco:
Fixed in https://github.com/Jc2k/aiohomekit/pull/43 and verified the traffic on the wire.


I noticed all the queriers get synchronized so they all try to refresh the A record at the same, but thats really a problem with zeroconf itself.  
https://datatracker.ietf.org/doc/html/rfc6762#section-5.2
```
   To avoid accidental synchronization when, for some reason, multiple
   clients begin querying at exactly the same moment (e.g., because of
   some common external trigger event), a Multicast DNS querier SHOULD
   also delay the first query of the series by a randomly chosen amount
   in the range 20-120 ms.
```

I'm off to bed.  I'll pick this back up in the morning.

Jc2k:
Good one. I'll put out a tag with these 2 fixes in this am.

Jc2k:
Both released in https://github.com/Jc2k/aiohomekit/releases/tag/0.4.1

PR #44, opened 2021-07-05
Ensure the correct zeroconf object is passed to AsyncServiceBrowser

bdraco:
The async version was being passed instead of the underlying version 🤦 

When I was doing development it worked this way, but I changed it before release for backwards compat and forgot to fix this when I updated the PR, and it just happened to work because MagicMock was a bit too magic and the devices I were testing with were already in the cache.

### DIFF
--- a/aiohomekit/zeroconf/__init__.py
+++ b/aiohomekit/zeroconf/__init__.py
@@ -277,7 +277,7 @@ async def _async_find_data_for_device_id(
     listener = CollectingListener(
         device_id=device_id, found_device_event=found_device_event
     )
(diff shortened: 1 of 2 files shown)


bdraco:
Its finding my myq bridge after I manually changed the `AccessoryIP` value in `.store/core.config_entries`.  However the ip is never updated so it always has to look it up at startup.  It think that needs to be fixed on the Home Assistant side though

Jc2k (approved):

Jc2k:
Will push out release when back at desk

PR #46, opened 2021-07-08
Use commentjson when parsing JSON from devices, as it may contain trailing commas

Jc2k:
See https://github.com/home-assistant/core/issues/52642


bdraco:
Tested and working as expected with multiple accessory and bridges

bdraco (approved):

Issue #1, opened 2020-03-22
Unknown parser state with homebridge as accessory


vilhalmer:
I just upgraded to hass 0.107.5, and during startup I get the following traceback:
```

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/asyncio/selector_events.py", line 826, in _read_ready__data_received
    self._protocol.data_received(data)
  File "/usr/local/lib/python3.7/site-packages/aiohomekit/controller/ip/connection.py", line 185, in data_received
    super().data_received(decrypted)
  File "/usr/local/lib/python3.7/site-packages/aiohomekit/controller/ip/connection.py", line 84, in data_received
    data = self.current_response.parse(data)
  File "/usr/local/lib/python3.7/site-packages/aiohomekit/http/response.py", line 95, in parse
    raise HttpException("Unknown parser state")
aiohomekit.exceptions.HttpException: Unknown parser state
```

I took a look at the parser and it seems like the only way this can happen is if more data arrives after `STATE_DONE` is set.

The only accessory I'm using is actually homebridge, to use another layer of translation that hass can't do natively yet. It's definitely possible that homebridge does something strange compared to a real accessory. I attempted re-pairing just to verify that homebridge is the culprit, and indeed the traceback appeared immediately.

Let me know what info I can provide to help debug this!

Jc2k:
Congrats on issue #1 ;) Homebridge has been a bit hit and miss so far. It does some stuff I know real HomeKit devices dont (like sending empty events as keep alives even when you havent turned push on at all). Did have a report of it working earlier today though.

I need you to modify your local copy of `aiohomekit/http/response.py` so that we can see whats in the buffer. Maybe something like this? https://gist.githubusercontent.com/Jc2k/c89e30fdc112b2329b12609bff8a9e07/raw/bee6b2950da45087addc7a8ffec5e3e1799e0a48/response.py

vilhalmer:
Sorry for the delay, it has been a thorougly weird week. I'll get this applied and report back,

vilhalmer:
Is this the right URL? There's no diff between this and the version on master.

Jc2k:
Oops! Lets try again. I think I mean [this one](https://gist.githubusercontent.com/Jc2k/c89e30fdc112b2329b12609bff8a9e07/raw/f996483fa184787c47e277e48ad5b72c08177f45/response.py) but it's been a few days so I can't really remember what i was thinking...

[...]

PR #47, opened 2021-07-11
Abstract json loading into aiohomekit.hkjson

bdraco:
- We try to use json.loads first and fallback to commentjson
  if we get a decode error to ensure that only devices that
  have the trailing commas have to pay the performance hit.

### DIFF
--- a/aiohomekit/controller/ip/connection.py
+++ b/aiohomekit/controller/ip/connection.py
@@ -18,8 +18,6 @@ import asyncio
 import json
 import logging
 
-import commentjson
-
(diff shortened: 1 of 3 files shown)


bdraco:
Watched during startup.  json loading went from #1 cpu user to so far down the list it doesn't matter.

bdraco:
Thankfully this turned out to be a simple change

Jc2k:
I think i'm broadly OK with this approach - hopefully we never have a device with lots of accessories and invalid JSON...!

PR #50, opened 2021-12-15
Added characteristic_types for ConnectSense

ecoen66:
I've added characteristic_types for ConnectSense SmartOutlet2 and In-Wall Outlets.


Jc2k (approved):
Brilliant. So there's no need for tests for this on the aiohomekit side. There's no way to check any more than the syntax, which the existing lints will do.

So if you are happy I'll merge this and tag, then we can get to work on the HA side.

With the tag out we can bump the dependency in the manifest and it's a lot easier than trying to have aiohomekit as a development checkout.

Are you comfortable with homekit_controller side of this? I can link to some example Prs or talk through anything your not sure on.

ecoen66:
I’ll give it a go.  Any examples would be great.
I’ve also got a few homebridge plugins that I wrote that I am going to want to recreate in HA.
Thanks!Ed


Sent from Yahoo Mail for iPhone


On Wednesday, December 15, 2021, 1:57 PM, Jc2k ***@***.***> wrote:




@Jc2k approved this pull request.

Brilliant. So there's no need for tests for this on the aiohomekit side. There's no way to check any more than the syntax, which the existing lints will do.

So if you are happy I'll merge this and tag, then we can get to work on the HA side.

With the tag out we can bump the dependency in the manifest and it's a lot easier than trying to have aiohomekit as a development checkout.

Are you comfortable with homekit_controller side of this? I can link to some example Prs or talk through anything your not sure on.

ecoen66:
lint/black wasn't happy with a trailing space on a comment line... I cleaned it up.

Jc2k:
So take a look at the changes to `sensor.py` in here: https://github.com/home-assistant/core/pull/54673/files.

The syntax has changed a bit on `dev` as its now using a typed EntityDescription rather than a dictionary, but the idea is the same.

Possibly an even better example is this one: https://github.com/home-assistant/core/pull/53891/files

It shows how to bump the aiohomekit version (need to update some requirements files as well as the manifest), what you need to add to `const.py` AND `sensor.py`. And how to add a test that proves JSON captures from an actual device is correctly enumerated and the right entities produced.

When you have updated the manifest file, running `./script/gen_requirements_all.py` should fix the other requirements files btw.

I've just pushed 0.6.6 to git, it should spit out a release on pypi in the next few minutes.

Jc2k:
@ecoen66 how are you getting on with this? I've just release 0.6.8, and here is the PR to update Home Assistant.

https://github.com/home-assistant/core/pull/62538/files

That should mean for your new sensors it will be enough to update sensor.py and const.py.

ecoen66:
Not great.  I fumbled a lot with the Dev environment before abandoning it and hacking the code on my production Pi for testing.  Forgot to disable updates and lost that code to a Core update.




I’m going to try the Dev container again instead, but couldn’t get it to discover anything.

Jc2k:
I don't use the Dev container, but what I do do (and might help) is I pair the device on my production instance then copy the config entry from `.storage/core.config_entries.yaml` to my dev environment. That looks something like this:

```
            {
                "entry_id": "efac887ca105d4124a2e4392948f3513",
                "version": 1,
                "domain": "homekit_controller",
                "title": "Aqara-Hub-E1-91C7",
                "data": {
                    "AccessoryPairingID": "xx:xx:xx:xx:xx:xx",
                    "AccessoryLTPK": "xxxxx",
                    "iOSPairingId": "xxxx",
                    "iOSDeviceLTSK": "xxxx",
                    "iOSDeviceLTPK": "xxxx",
                    "AccessoryIP": "1.2.3.4",
                    "AccessoryPort": 42553,
                    "Connection": "IP"
                },
                "options": {},
                "pref_disable_new_entities": false,
                "pref_disable_polling": false,
                "source": "zeroconf",
                "unique_id": "x:x:x:x:x:x",
                "disabled_by": null
            }
```

It will recreate everything else it needs (like the entity map file) automatically.

This might help you because discovery doesn't strictly need to work as long as the pairing exists and the IP and port are correct.

This will also mean you can pair with production and dev at the same time (by sharing keys). Rather than having to keep unpairing (normally a device is undiscoverable if its already paired with something else).

Jc2k:
Typo in original reply fixed - it's `.storage/core.config_entries` you need.

Jc2k:
If you go down this route, be careful not to delete the pairing in your dev instance through the UI - i think (can't remember if i got the PR merged) it will delete the pairing from the device and break your prod instance as well.

ecoen66:
Working now!  Thanks.  Now I just have to submit a pull request for core.

Jc2k:
Great stuff! The bots should automatically tag me when they see you touched homekit_controller, so I'll watch out for it!

ecoen66:
pull request is in for core:dev

PR #51, opened 2022-01-15
Fix step value

roysjosh:
Without this, the HAP/CoAP Nanoleaf Essentials bulb was being sent
hue/sat values with high precision, against the spec step value.


roysjosh:
The bulbs don't specify step value (or min or max etc) for any characteristic so they'll rely on the builtin definitions

Jc2k:
I think i'd rather fix `data.py` to use `min_step` (there was a typo in `scripts/generate_metadata.py` so that file is wrong).

Jc2k (approved):

Jc2k:
Cheers!

PR #52, opened 2022-01-15
HAP over CoAP

roysjosh:
WIP, looking for initial feedback

### DIFF
--- a/aiohomekit/controller/coap/connection.py
+++ b/aiohomekit/controller/coap/connection.py
@@ -1,0 +1,725 @@
+#
+# Copyright 2022 aiohomekit team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from __future__ import annotations
+
+import asyncio
+import logging
+import random
+import struct
+from typing import Any
+import uuid
+
+from aiocoap import Context, Message, resource
+from aiocoap.error import NetworkError
+from aiocoap.numbers.codes import Code
+from cryptography.exceptions import InvalidTag
+from cryptography.hazmat.primitives.ciphers.aead import ChaCha20Poly1305
+
+from aiohomekit.exceptions import (
+    AccessoryDisconnectedError,
+    AuthenticationError,
+    EncryptionError,
+    UnknownError,
+)
+from aiohomekit.protocol import (
+    get_session_keys,
+    perform_pair_setup_part1,
+    perform_pair_setup_part2,
+)
+from aiohomekit.protocol.tlv import HAP_TLV, TLV
+
+from .pdu import OpCode, PDUStatus, decode_pdu, encode_all_pdus, encode_pdu
+from .structs import Pdu09Database
+
+logger = logging.getLogger(__name__)
+
+
+def decode_pdu_03(buf):
+    return bytes(dict(TLV.decode_bytes(buf)).get(HAP_TLV.kTLVHAPParamValue))
+
+
+def decode_list_pairings_response(buf):
+    inner_bytes = decode_pdu_03(buf)
+    return TLV.decode_bytes(inner_bytes)
+
+
+class EncryptionContext:
+
+    coap_ctx: Context
+    lock: asyncio.Lock
+    uri: str
+
+    event_ctr: int
+    event_ctx: ChaCha20Poly1305
+    recv_ctr: int
+    recv_ctx: ChaCha20Poly1305
+    send_ctr: int
+    send_ctx: ChaCha20Poly1305
+
+    def __init__(self, recv_ctx, send_ctx, event_ctx, uri, coap_ctx):
+        self.recv_ctr = 0
+        self.recv_ctx = recv_ctx
+        self.send_ctr = 0
+        self.send_ctx = send_ctx
+        self.event_ctr = 0
+        self.event_ctx = event_ctx
+
+        self.coap_ctx = coap_ctx
+        self.lock = asyncio.Lock()
+        self.uri = uri
+
+    def decrypt(self, enc_data: bytes) -> bytes:
+        logger.debug("DECRYPT counter=%d" % (self.recv_ctr,))
+        dec_data = self.recv_ctx.decrypt(
+            struct.pack("=4xQ", self.recv_ctr), enc_data, b""
+        )
+        self.recv_ctr += 1
+        return dec_data
+
+    def decrypt_event(self, enc_data: bytes) -> bytes:
+        dec_data = self.event_ctx.decrypt(
+            struct.pack("=4xQ", self.event_ctr), enc_data, b""
+        )
+        self.event_ctr += 1
+        return dec_data
+
+    def encrypt(self, dec_data: bytes) -> bytes:
+        logger.debug("ENCRYPT counter=%d" % (self.send_ctr,))
+        enc_data = self.send_ctx.encrypt(
+            struct.pack("=4xQ", self.send_ctr), dec_data, b""
+        )
+        self.send_ctr += 1
+        return enc_data
+
+    async def _decrypt_response(self, response: Message):
+        try:
+            return self.decrypt(response.payload)
+        except InvalidTag:
+            logger.error(
+                "Decryption failed, desynchronized? Counter=%d/%d"
+                % (self.recv_ctr, self.send_ctr)
+            )
+
+            # look back a few counter values
+            rewind = min(5, self.recv_ctr)
+            self.recv_ctr -= rewind
+            for i in range(rewind):
+                logger.debug("Attempting to recover by rewind, try %d" % (i + 1,))
+                try:
+                    return self.decrypt(response.payload)
+                except InvalidTag:
+                    self.recv_ctr += 1
+
+            # fast forward a few counter values
+            for i in range(5):
+                logger.debug("Attempting to recover, try %d" % (i + 1,))
+                try:
+                    # attempt to resynchronize by moving the counter forward
+                    # we've got to roll it forward ourselves as the exception prevents that
+                    self.recv_ctr += 1
+                    return self.decrypt(response.payload)
+                except InvalidTag:
+                    pass
+
+            # try zeroing out the counters
+            try:
+                self.recv_ctr = 0
+                self.send_ctr = 0
+                return self.decrypt(response.payload)
+            except InvalidTag:
+                pass
+
+            logger.error(
+                "Failed flailing attempts to resynchronize, self-destructing in 3, 2, 1..."
+            )
+
+            await self.coap_ctx.shutdown()
+            self.coap_ctx = None
+            raise EncryptionError("Decryption of PDU POST response failed")
+
+    async def post_bytes(self, payload: bytes, timeout: int = 16.0):
+        async with self.lock:
+            payload = self.encrypt(payload)
+
+            try:
+                request = Message(code=Code.POST, payload=payload, uri=self.uri)
+                response = await asyncio.wait_for(
+                    self.coap_ctx.request(request).response, timeout=timeout
+                )
+            except (NetworkError, asyncio.TimeoutError):
+                raise AccessoryDisconnectedError("Request timeout")
+
+            if response.code != Code.CHANGED:
+                logger.warning(f"CoAP POST returned unexpected code {response}")
+
+            return await self._decrypt_response(response)
+
+    async def post(
+        self, opcode: OpCode, iid: int, data: bytes
+    ) -> tuple[int, bytes | PDUStatus | ValueError]:
+        tid = random.randint(1, 254)
+        req_pdu = encode_pdu(opcode, tid, iid, data)
+        res_pdu = await self.post_bytes(req_pdu)
+        return decode_pdu(tid, res_pdu)
+
+    async def post_all(
+        self, opcode: OpCode, iids: list[int], data: list[bytes]
+    ) -> list[bytes | PDUStatus | ValueError]:
+        req_pdu = encode_all_pdus(opcode, iids, data)
+        res_pdu = await self.post_bytes(req_pdu)
+
+        idx = 0
+        offset = 0
+        res = []
+        while True:
+            body_len, body = decode_pdu(idx, res_pdu[offset:])
+            res.append(body)
+
+            idx += 1
+            offset += 5 + body_len
+            if offset >= len(res_pdu):
+                break
+
+        return res
+
+
+class EventResource(resource.Resource):
+    def __init__(self, connection):
+        super().__init__()
+        self.connection = connection
+
+    async def render_put(self, request):
+        try:
+            payload = self.connection.enc_ctx.decrypt_event(request.payload)
+        except InvalidTag:
+            logger.warning(
+                "Event decryption failed, desynchronized? Counter=%d"
+                % (self.connection.enc_ctx.event_ctr,)
+            )
+            # XXX invalidate subscriptions, etc
+            return Message(code=Code.NOT_FOUND)
+
+        logger.debug(f"CoAP event: {payload.hex()}")
+
+        offset = 0
+        while True:
+            _, iid, body_len = struct.unpack("<BHH", payload[offset : offset + 5])
+            body = payload[offset + 5 : offset + 5 + body_len]
+
+            characteristic = self.connection.info.find_characteristic_by_iid(iid)
+            value = decode_pdu_03(body) if body_len > 0 else b""
+            if characteristic is not None and body_len > 0:
+                characteristic.raw_value = value
+                value = characteristic.value
+            logger.debug("event ?/%d = %r" % (iid, value))
+
+            if self.connection.owner:
+                # XXX aid
+                key = (1, iid)
+                self.connection.owner.event_received(
+                    {
+                        key: {
+                            "value": value,
+                        }
+                    }
+                )
+
+            offset += 5 + body_len
+            if offset >= len(payload):
+                break
+
+        return Message(code=Code.VALID)
+
+
+class CoAPHomeKitConnection:
+    def __init__(self, owner, host, port):
+        self.address = f"[{host}]:{port}"
+        self.connection_lock = asyncio.Lock()
+        self.enc_ctx = None
+        self.owner = owner
+        self.pair_setup_client = None
+
+    async def reconnect_soon(self, updated_ip_port=None):
+        if updated_ip_port is None:
+            return
+        old_address = self.address
+        new_ip = updated_ip_port["AccessoryIP"]
+        new_port = updated_ip_port["AccessoryPort"]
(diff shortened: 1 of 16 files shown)


roysjosh:
Branch where HA work is being done: https://github.com/roysjosh/core/tree/hap-over-coap

Jc2k:
This is looking good. I'll try and find time to give it a quick review this weekend. The first thing that catches my eye is, can we share code for the pair-setup and pair-verify stuff with the existing pairing code? The existing code is already decoupled from TCP ("sans io") and has been used with BLE successfully, but can be refactored further if needed.

roysjosh:
It is very likely they can be combined. One of the few differences is the creation of a 3rd encryption context for events at the end of pair-verify. I'll find a way to combine them.

aiohomekit/controller/coap/connection.py
```diff
@@ -0,0 +1,971 @@
+
+import asyncio
+import struct
+import logging
+import tlv8
```
Jc2k, line 5: I'm guessing this is https://pypi.org/project/tlv8/ and not aiohomekit/aiohomekit/protocol/tlv.py?

`tlv8` was forked (and improved) from homekit_python. aiohomekit was also forked from homekit_python as my asyncio changes were not really a good fit upstream. But my fork predates the work on homekit_python to pull out tlv8. So here we are currently stuck with whats in aiohomekit/protocol/tlv.py from homekit_python.

We also have https://github.com/Jc2k/aiohomekit/blob/5902621ca0a1c8bd4399fcf8f2b7767d880cc252/aiohomekit/model/characteristics/structs.py for decoding complicated nested structs. You can do something like `SelectedRTPStreamConfiguration.decode(bytes)`. See `aiohomekit/tlv8.py`. 

Don't want to have 2 tlv8 libraries on the go at once, so for now better to stick with whats available in the code base. We can look at moving to tlv8 seperately if its worth it.

aiohomekit/controller/coap/connection.py
```diff
@@ -0,0 +1,971 @@
...
+from cryptography.hazmat.primitives.ciphers.aead import ChaCha20Poly1305
+
+from aiohomekit.crypto.hkdf import hkdf_derive
+from aiohomekit.crypto.srp import SrpClient
+
+
+# globals/enums
+logger = logging.getLogger(__name__)
+
+class HAP_PAIRING_ERRORS:
```
Jc2k, line 23: There are already constants for these, again inherited from homekit_python. They are not in the most useful format but we should try and consolidated (search for kTLVError_MaxPeers).

Jc2k: I'm also trying to use python enums for new things, as its more useful for type checking and code completion.

aiohomekit/controller/coap/connection.py
```diff
@@ -0,0 +1,971 @@
...
+    UNK_13_CHARACTERISTIC = 0x13
+    UNK_14_CHARACTERISTICS = 0x14
+    UNK_15_SERVICE = 0x15
+    UNK_16_SERVICES = 0x16
+    UNK_17 = 0x17
+    UNK_18 = 0x18
+    UNK_19 = 0x19
+    UNK_1A_ACCESSORY_INSTANCE_ID = 0x1A
+
+class HAP_TLV_TAGS:
```
Jc2k, line 82: Some crossover here with https://github.com/Jc2k/aiohomekit/blob/5902621ca0a1c8bd4399fcf8f2b7767d880cc252/aiohomekit/protocol/tlv.py#L44.

Jc2k (changes requested):
I really like the direction this is going in. I'll probably wait for you to de-dupe the crypto bits before I have another look.

In general try to follow Home Assistant's coding style and naming conventions and best practices (https://developers.home-assistant.io/docs/development_guidelines). Old code from homekit_python doesn't, though thats only until I have time to refactor it.

Go wild with dataclasses, enums and typing generally. A lot of the ones in aiohomekit were generated automatically and are rubbish. Feel free to fix as needed.

There is a lot of legacy cleanup in here to do still, so if something is annoying feel free to refactor it - especially if its an internal bit not exposed to Home Assistant (if it does impact Home Assistant, we'll need a PR to both). Any cleanups you want, can you put them in seperate PR's to keep this one clean.

roysjosh:
Hmm I need to find a way to mark the bulb as unavailable when the CoAP comms time out + invalidate the subscriptions whenever the CoAP session needs to be recreated (pair-verify re-executed) as that generates new push event keys.

Jc2k:
I've never done anything with CoAP. Is there a maximum age for a CoAP session? Or is it based on the time since the last exchange?

The super simple way would just be to use a `time.monotonic()` (so its DST safe) and have `ensure_connected()` sort out the session if the timer has expired. HA polls every 30s, so if the session drops out and we don't notice, then HA will bring it back to life in at most 30s. Ultimately this is what we are doing for TCP when the networking stack doesn't notice the connection drop (like a device crashes and reboots it won't RST existing connections).

That said, I would like to eventually be smarter about polling (it's pretty hard on the battery of my Eve Degree 😢). So alternatively you could start an asyncio task that will close the session after a timeout, and cancel it and reschedule it every time there is a succesful communication with the device? This is kind of annoying in asyncio because you'd need to spawn a coroutine, but then something needs to collect the results (or exceptions) of the coroutine, otherwise you leak expections and things can end up silently broken. I don't think we've got anything like that in aiohomekit atm that you can reuse, though.

Or did you have something nicer in mind?

roysjosh:
I'll squash all of this at the end. For now I'll try to commit chunks of enhancements under brief but descriptive commits to help you see what has changed. I'll also comment on your open review items. Big item in this batch was the crypto de-dupe which somehow went without incident. Yay for good specs.

aiohomekit/controller/coap/connection.py
roysjosh: Sounds good, I'll work to pull tlv8 out of this PR

roysjosh:
I'm guessing the CoAP behavior is going to be implementation-specific where the spec isn't clear. Sessions are likely tied to IP & port. I haven't left my HA dev instance up more than a half of a day yet, plus the NL Thread network occasionally has a hiccup which tends to reset the comms so I'm not sure if these bulbs have a maximum session lifetime yet. I'm aiming to make the crypto bits robust to desynchronization or full context resets. It's a lot better than the first code but likely still more to go.

aiohomekit/controller/coap/connection.py
```diff
@@ -0,0 +1,830 @@
...
+            except NetworkError:
+                raise AccessoryDisconnectedError('Request timeout')
+
+            if response.code != Code.CHANGED:
+                logger.warn('CoAP POST returned unexpected code %s' % (response.code,))
+
+            try:
+                payload = self.encCtx.decrypt(response.payload)
+            except InvalidTag as ex:
+                logger.error('Decryption failed, desynchronized? Counter=%d/%d' % (self.encCtx.recvCtr, self.encCtx.sendCtr))
```
Jc2k, line 614: Worth pulling this into `EncryptionContext`? Easier to test there as you don't have to mock any i/o, and avoids leaking `recvCtr` awareness into your connection code.

Jc2k:
> . I'll also comment on your open review items. Big item in this batch was the crypto de-dupe which somehow went without incident. Yay for good specs.

Sounds good!

pyproject.toml
```diff
@@ -37,6 +37,7 @@ pyupgrade = "^2.7.2"
 pytest-cov = "^2.10.1"
 asynctest = "^0.13.0"
 aiohttp = "^3.8.1"
+aiocoap = "^0.4.1"
```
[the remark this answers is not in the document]
roysjosh: Updated, although many other deps were also touched.

aiohomekit/controller/coap/connection.py
roysjosh: Removed errors & methods as they aren't needed now that crypto is merged.

aiohomekit/controller/coap/connection.py
roysjosh: Now using an IntEnum in tlv.py for HAP TLV tags.

aiohomekit/controller/coap/connection.py
roysjosh: I moved the post function to the encryption context class. I'll probably do a little more refactoring here. I don't like that the device address is retained in two different places.

aiohomekit/controller/coap/connection.py
roysjosh: tlv8 has been removed

roysjosh:
@Jc2k okay, I think this is ready for another round of reviews, please. I've had my dev HA up for most of 3 days without any major hiccups.

roysjosh:
RE: polling, do you have any feel for if devices generally support events properly? iOS very rarely polled my fake bulb app's `lightbulb` characteristics & seemed to largely rely on events. It did poll the `ping` characteristic under the `accessory runtime information` service on a regular but infrequent basis.

roysjosh:
I've played around with the Home app some more as well as the Eve app. It seems like every time an app needs to see if the accessory is still alive it reads the `ping` endpoint. If it gets an error response, it does a pair-verify to reconnect. No response makes the accessory display "Updating..." until it responds. Any time the app has to reconnect it re-reads all the characteristics. But if the initial `ping` comes back, it seems like the assumption is that the state shouldn't have changed since things are working normally and it doesn't re-read the primary service's characteristics.

Jc2k:
RE: Polling and events. Good question. We don't have any telemetry so I don't have any solid answers, here are some observations.

Regardless of reliability, I know some characteristics are poll only. These days this is less of a problem but I think in R1 of the HAP spec maybe temperature was poll only (it has events these days for sure).

I know we have some TCP HAP devices that refuse subscription requests, so we actually have a path that disables events if it detects this has happened. (It toasts the secure session, so we have to stop trying to even send the subscription request). This is probably a bug on our end but it's not been a big enough problem to look into further yet.

I know we have some devices that either have really flaky WiFi or they are just on the edge of their owners range. So we'd get packet loss related failures often. This has caused lots of tickets which are mostly just more errors to silence.

Connections dropping without a TCP RST was common. This includes from devices crashing.

IIRC one of my devices seems to have a maximum TCP uptime of about 17 hours before the connection needs rotating.

While not about events reliability exactly we have observed that the implementations vary wildly in quality. Some favourite bugs:

* A device could parse `{"foo":"bar"}` but not `{"foo": "bar}` (yes, the space)
* A device could generate invalid JSON that python can't parse (`["foo",]`)
* A device could handle `1`, `0` and `true` for a **BOOL** field, but not `false`.
* A device was really picky about the structure of its `Host` header
* A device was sensitive about how many packets were used to send HTTP requests
* A device randomly returns a 500 or 400 with HTML instead of HAP response for some requests
* Devices returning e.g. 70402 instead of -70402

Which has made me very mistrustful of them.

All that said, the events implementation on Hue has been really solid. I have a Hue Remote that i use in my office (i.e. multiple times daily for a year) and i've never noticed it drop an event. Just got an outdoor Friends of Hue switch that so far that is also pretty solid. 

I would like to rely on events more in the future. I have an Eve Degree and it eats batteries because of the polling. But I probably won't be just switching off polling completely given the vast range of devices we are dealing with. It would be good to slowly back off the poll rate if events are seen to be working for a connection maybe. But with some sort of ping to stop the TCP connection timing out. I don't think the `ping` characteristic is widely available, although it could be i just didn't notice it before. So i'd probably just poll something else like name.

Jc2k:
RE: Review - i'll make it my next priority, though might be next week now.

How easy would it be for you to pull some bits out into seperate PR's? E.g. the changes to tlv8./tlv.py would be a good candidate, they are quite self contained and we can easily add tests for them. (It's quite large, so if we can get the easy bits landed then there is less noise when im re-reviewing). This is what I do to help the Home Assistant reviewers with my larger changes upstream, its a bit of a PITA with all the rebasing but a big help. And i reckon things like the tlv changes on their own I can review sooner.

Jc2k:
I'd like to look at pulling out the TLV structs into `aiohomekit/controller/coap/structs.py` and add some tests for them (ideally decoding actually in-the-wild PDU data as the next thing when the outstanding PR's are merged.

I'd also like to see if we can reuse `.model.Accessories`, `.model.Service` and `.model.Characteristic`. They already have code to synthesize the IP-style JSON responses and they have a layer to do lookups by iid and type. I'm currently trying to make aiohomekit nicer to use from home assistant and i'm thinking of having `list_accessories_and_characteristics` return an `Accessories` as I just load it into one by hand on the other side anyway. So it would be good to use them from the start.

I'm sort of imagining `Pdu09Characteristic` gains a `to_characteristic`, `Pdu09Service` gains a `to_service`, `Pdu09Accessory` gains a `to_accessory` and `Pdu09Database` gains `to_accessories`.

aiohomekit/controller/coap/pairing.py
```diff
@@ -0,0 +1,143 @@
...
+
+    async def list_accessories_and_characteristics(self):
+        await self._ensure_connected()
+
+        accessories = await self.connection.get_accessory_info()
+
+        for accessory in accessories:
+            for service in accessory["services"]:
+                try:
+                    service["type"] = ServicesTypes.get_uuid(service["type"])
```
Jc2k, line 81: If you rebase this is gone in favour of `aiohomekit.uuid.normalize_uuid`. It won't raise on unknown UUID's so you can lose the KeyError too.

aiohomekit/controller/coap/pairing.py
```diff
@@ -0,0 +1,143 @@
...
+                except KeyError:
+                    pass
+
+                for characteristic in service["characteristics"]:
+                    try:
+                        characteristic["type"] = CharacteristicsTypes.get_uuid(
```
Jc2k, line 87: This is also gone in favour of `aiohomekit.uuid.normalize_uuid`. You can lose the KeyError there too.

aiohomekit/controller/coap/connection.py
```diff
@@ -0,0 +1,1097 @@
...
+                response = await asyncio.wait_for(
+                    self.coapCtx.request(request).response, timeout=timeout
+                )
+            except (NetworkError, asyncio.TimeoutError):
+                raise AccessoryDisconnectedError("Request timeout")
+
+            if response.code != Code.CHANGED:
+                logger.warn(f"CoAP POST returned unexpected code {response}")
+
+            try:
```
Jc2k, line 468: Theres no i/o in the decrypt retry code. So can we pull it out of `post` so thats its more testable, and add some tests for it.

Jc2k (commented):
Also in parallel, you could look at getting the `parsed_address` change made in home assistant zeroconf. I'm not going to be the one reviewing that and you might as well get thee ball rolling.

Jc2k:
RE: Tests, IP currently has a mock accessory that it uses. I hate this approach. It's slow and fails at random. Open to ideas on how you'd like to proceed there. I don't want to drown you in test writing, but it would be good to have some coverage of the happy path as I am in the midst of a (admittedly quite slow) cleanup exercise and could very easily break this.

We could break the i/o out, so for `read_characteristics` we might have:

```python
def serialize_read_characteristics_request(ids: list[tuple[int, int]]) -> bytes:
    pdu = b"".join(
        [
            hap_pdu(
                HAP_PDU_OPCODES.HAP_CHARACTERISTIC_READ, idx, int(aid_iid[1]), b""
            )
            for (idx, aid_iid) in enumerate(ids)
        ]
    )

def deserialize_read_characteristics_response(ids: list[tuple[int, int]], payload: bytes) -> dict[str, Any]:
        offset = 0
        results = dict()
        while True:
            pduControl, pduTid, pduStatus, pduBodyLen = struct.unpack(
                "<BBBH", payload[offset : offset + 5]
            )
            aid_iid = ids[pduTid]
            logger.debug(
                "Read value for aid %d iid %d: %s, len %d"
                % (
                    int(aid_iid[0]),
                    int(aid_iid[1]),
                    pduStatusMap[pduStatus],
                    pduBodyLen,
                )
            )

            if pduStatus == 0:
                # decode TLV to get byte value
                value = (
                    decode_pdu_03(payload[offset + 5 : offset + 5 + pduBodyLen])
                    if pduBodyLen > 0
                    else b""
                )
                # find characteristic so we can get the data type
                characteristic = self.info.find_characteristic_by_iid(int(aid_iid[1]))
                # if we found it & have a non-empty value...
                if characteristic is not None and pduBodyLen > 0:
                    # set the raw bytes
                    characteristic.raw_value = value
                    # and get the decoded value
                    value = characteristic.value
                # add result to dict
                results[aid_iid] = {
                    "value": value,
                }
            else:
                results[aid_iid] = {
                    "description": pduStatusMap[pduStatus],
                    "status": pduStatus,
                }

            offset += 5 + pduBodyLen
            if offset >= len(payload):
                break

    async def read_characteristics(self, ids):
        pdu = serialize_read_characteristics_request(ids)
        payload = await self.encCtx.post(pdu)
        results = deserialize_read_characteristics_response(payload)
        logger.debug(f"Read characteristics: {results!r}")
        return results
```

Both of the 2 helpers are "pure", they don't need to know about the connection classes or any of the state on them. They could even live in a different module. So they are super easy to test.

We could also choose to mock `self.encCtx.post`. My only thought there is I'd like to have too many mocks, as they can be quite brittle when refactoring. So i'd probably lean towards the first approach.

roysjosh:
Rebased w/o changes, more commits on the way to address comments

Jc2k:
Awesome. Your work has inspired me to sort BLE out (had it working in homekit_controller aaages ago, but ripped it out when I did the asyncio port). So i'm expecting to be able to reuse bits of this there.

Jc2k:
(Sorry for close, child helping 😵‍💫 )

roysjosh:
BLE would be great! One day we could set up devices completely in HA.

Jc2k:
I've got pairing and get/put/identify/list-pairings working already via the CLI. Just cleaning it up a bit and adding some tests. `bleak` is so much nicer to use than when we did this with dbus-python for homekit_python!

Not sure i'll add events straight away, there are 2 different event systems (connected and disconnected) and i need to have a background task constantly monitoring the advertisement data to pick up disconnected events. I think I want to clean up aiohomekit a lot more first.

Hopefully we can find something in the Thread Control Point for managing the thread network.

Draft is here: https://github.com/Jc2k/aiohomekit/pull/60.

Jc2k:
Would it be OK with you if we made a `dev` branch and we put both BLE and CoAP in there, and both work against the same branch for a bit? We'd both raise PR's against it till we were happy with it.

I want to rework the discovery interface a bit, and I need to change all 3 backends to do it, so will be easier this way and you can still work on the other bits in parallel.

roysjosh:
I think that makes sense. I'm refactoring a few things still & then I'll work on some tests.

Jc2k:
Good one. It'll do it tomorrow then (hopefully).

roysjosh:
Squashed this PR

Jc2k:
`dev` branch is open with this and BLE support is merged. Let's work there, with PR's per cleanup/fix/feature. When ready this will become 1.0.0 because its such a huge new set of features.

I'll open up tickets with a coap label in the 1.0.0 milestone and probably tag you as well where there is something I want to change or clarify. Feel free to raise things in 1.0.0 either to ask me or just to keep track of our progress.

I'll open PR's and assign them to you where I touch coap stuff, at least while we are both working on this.

My area of focus is going to be around zeroconf and BLE discovery for the time being. One thing I want to change is the weird lifecycle zeroconf has:

* In HA, an existing zeroconf instance is passed to us. It's cache has already made a bunch of discoveries.
* In the CLI, the Controller has to make its own zeroconf instance. It's completely cold.

It's making that code a bit harder to reason about.

Cheers for all the work you've done on this. Hopefully nearly there with it..

PR #55, opened 2022-02-01
Return a key derivation function as a result of pair-verify

roysjosh:
This will allow other keys beyond the current two to be derived.


Jc2k (approved):
LGTM. Happy that this is covered by existing tests.

PR #56, opened 2022-02-01
Add u64, u128, and bytes to tlv8

roysjosh:


Jc2k:
Could you add something to https://github.com/Jc2k/aiohomekit/blob/main/tests/test_tlv8.py to cover these? Otherwise i'm happy with this.

Jc2k (approved):

PR #57, opened 2022-02-01
Ignore tlv8 dataclass fields marked with init=False

roysjosh:
This allows dataclasses to have supporting fields.


Jc2k:
Happy with this, but again can we add a test to https://github.com/Jc2k/aiohomekit/blob/main/tests/test_tlv8.py.

Jc2k (approved):

PR #60, opened 2022-02-03
HAP over BLE

Jc2k:
This implements BLE support using bleak.

Todo:

* [ ] Tests
* [ ] Error handling, especially for put / get characteristic.
* [x] Add pairing
* [x] Remove pairing

Low priority:

* [ ] Connected events.
* [ ] Disconnected events. Needs a background task to monitor BLE advertisements. Needs to trigger a poll of every subscribed characteristic, and then send events through existing mechanism.
* [ ] Characteristic cache refresh when c# increment happens - this is slightly lower priority than with IP devices. Given we own the admin slot and don't do firmware updates the database is much less likely to get out of date. Can still happen though for devices offering firmware updates via other protocols. Need a backround task to monitor BLE advertisements.


Jc2k:
@los93sol hey - wanted to give you a heads up about this PR. HomeKit BLE might be coming to Home Assistant sooner rather than later!

I also wanted to thank you for the tip about ATT_MTU and about the Pairing Features characteristic. Both should be handled in this branch.

los93sol:
Nice!  I’ve been working on a project that exposes BLE devices as IP and have it working currently with Home Assistant.  Still a few stability issues to work through but it’s working with Level locks.  I think both are good since my solution can be installed on a pi and dropped near the device so it helps address scenarios where your hass machine either doesn’t have Bluetooth or where it’s out of range.

los93sol:
@Jc2k Another thing to watch for is characteristics with tw permission…seems to mean that timed write is required.  On this Level lock I have, I had to do the timed write procedure to lock and unlock it.  Let me know if you need any additional info about it, the only real trick is sending a timed write then immediately sending an execute

Jc2k:
Cheers @los93sol - latest commit should add that.

Jc2k:
This has landed on `dev` (what will become 1.0.0).

PR #65, opened 2022-02-08
Discovery refactor

Jc2k:
This is still a WIP.

The main changes are:

* The controller can run background services. This is used to e.g. run the BLE scanner in the background.
* The main controller is now just a meta layer over 3 protocol specific controllers. This is really just thinking about the future - we are going to end up with devices that have multiple transports. IP > CoAP > BLE. So the end game is I want a Pairing that can switch between transports (like iOS apparently does). But for now, I just consult the transports in priority order and pick the best match.
* `discover_ip` and `discover_ble` go in favour of `async_discover` which works for ALL backends.
* `find_ip_by_device_id` goes in favour of `async_find` which works for ALL backends.

The typing is much better, and the discovery fields are accessed on the Discovery dict directly rather than fishing around in an untyped dictionary.

The naming is just for consistency with HA (where `async_` means safe to call from an async context, regardless of whether its awaitable or not).


Jc2k:
HA branch: https://github.com/home-assistant/core/compare/dev...Jc2k:homekit_controller_coap_ble?expand=1

Jc2k:
@roysjosh This is probably in a goodish place to land on dev now. A single `aiohomekitctl discover` can see both BLE and IP devices here, and AFAICT will see CoAP devices too.

With my [HA branch](https://github.com/home-assistant/core/compare/dev...Jc2k:homekit_controller_coap_ble?expand=1), I was able to pair a BLE device (using the "Add integration" route). I believe the abstractions in place mean HA won't know (or care about) the difference between an IP/CoAP backed pairing/discovery, so all 3 should work.

When i've merged (probably tomorrow AM unless you yell) it I'm going to:

* Try and avoid homekit_controller assuming a discovery and a pairing have anything to do with zeroconf
* Deep-dive into `AsyncZeroconf`. I want to keep `aiohomekitctl` as close to HA as possible, to streamline our code around it. Right now i'm not entirely sure whether its considered OK to create an `AsyncServiceBrowser` for `_hap._tcp.local` if HA has already created one. API wise i'm pretty sure it works. But does it lead to extra traffic?
* Do some more end to end tests (unpairing, re-pairing, etc). I think BLE will need some robustifying, which I don't want to do in this PR.
* Tests. I'm going to split BLE with the same style you used and then add some tests.

PR #80, opened 2022-03-01
Add missing await

balloob:
There was an await missing when calling `discover_ip`


Jc2k:
Cheers!

Jc2k:
I can't release from my phone but it will be out by 10 am uk time tomorrow.

PR #84, opened 2022-04-19
Fix zeroconf lookups when the device id is UPPER case

bdraco:
F i x e s #https://github.com/home-assistant/core/issues/67665

### DIFF
--- a/aiohomekit/zeroconf/__init__.py
+++ b/aiohomekit/zeroconf/__init__.py
@@ -23,7 +23,7 @@ import contextlib
 import logging
 from typing import Any, Callable
 
-from zeroconf import ServiceBrowser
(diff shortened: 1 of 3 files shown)


aiohomekit/zeroconf/__init__.py
```diff
@@ -285,9 +294,7 @@ async def _async_find_data_for_device_id(
 
     try:
         for info in listener.get_data():
-            if not _service_info_is_homekit_device(info):
```
bdraco, line 288: I took this out since we are already filtering on HAP_TYPE above and we are looking for a specific id match below

Jc2k (approved):

bdraco:
@Jc2k would you mind pushing a new release?

bdraco:
Thanks!

PR #87, opened 2022-06-19
Use ChaCha20Poly1305Reusable for encryption

bdraco:
I built this for the other side of the house since I was annoyed at how long it took my phone to switch over from Mobile to WiFi and wanted to decrease the latency. It also seems to fix the Home app stalling out on iOS 16 betas when using them with Home Assistant's HomeKit integration.

- ChaCha20Poly1305 recreates the ctx every time its called which
  is an expensive operation. This makes quite a bit of difference
  when encrypting large blocks of data (camera images) since each
  block had to create a new context. cryptography does this because
  it needs calls to ChaCha20Poly1305 to be threadsafe at the cost
  of performance. Since we are using asyncio we can avoid the
  overhead of recreating the context every time since the encryptor
  and decryptor are never called from different threads

### DIFF
--- a/aiohomekit/controller/ip/connection.py
+++ b/aiohomekit/controller/ip/connection.py
@@ -19,8 +19,8 @@ import json
 import logging
 
 from aiohomekit.crypto.chacha20poly1305 import (
-    chacha20_aead_decrypt,
-    chacha20_aead_encrypt,
+    ChaCha20Poly1305Decryptor,
+    ChaCha20Poly1305Encryptor,
 )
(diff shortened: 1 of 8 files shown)


Jc2k:
Really happy with the change. Will probably need your help making sure I don't screw up the merge into dev. (Dev is where the thread and BLE support is languishing).

Jc2k:
Ah I misremembered. Dev changes some of the key generation stuff to make it easier to share with the thread code but it must have been the x25519 I'm thinking of. No worries then.

PR #94, opened 2022-07-06
Change chacha20poly1305-reuseable target to >=

bdraco:
- There was a [small typing fix ](https://github.com/bdraco/chacha20poly1305-reuseable/compare/v0.0.3...v0.0.4)for the lib, and I realized
  I should have added it with >=

Fixes
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
aiohomekit 0.7.19 requires chacha20poly1305-reuseable<0.0.4,>=0.0.3, but you have chacha20poly1305-reuseable 0.0.4 which is incompatible.
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -123,7 +123,7 @@ pycparser = "*"
 
 [[package]]
 name = "chacha20poly1305-reuseable"
-version = "0.0.3"
+version = "0.0.4"
 description = "ChaCha20Poly1305 that is reuseable for asyncio"
 category = "main"
 optional = false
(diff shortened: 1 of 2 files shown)


bdraco:
I'd like to bump this in HASS so we ship 0.0.4 with 2022.7

I can do the release if it is easier.

bdraco:
I figure you are probably still half asleep so I went ahead and pushed it.  The CI threw an error but it seems to have published ok.

PR #105, opened 2022-07-16
Auto discover the MTU

bdraco:


aiohomekit/controller/ble/client.py
```diff
@@ -213,3 +214,29 @@ async def drive_pairing_state_machine(
             request, expected = state_machine.send(TLV.decode_bytes(response))
         except StopIteration as result:
             return result.value
+
+
+class AIOHomeKitBleakClient(BleakClient):
```
bdraco, line 219: This should also give us a pretty decent patch target as well

Jc2k (approved):

bdraco:
Retesting the nanoleaf with the auto recover for finish pairing removed

PR #107, opened 2022-07-16
Fix BLE failing to write on M1 macs

bdraco:
See https://github.com/hbldh/bleak/issues/837


Jc2k (approved):

bdraco:
Linux still works with these changes

PR #114, opened 2022-07-17
Add is_available and transport properties

bdraco:
Also fixes a stall on startup where connect does not actually timeout


aiohomekit/controller/ble/pairing.py
```diff
@@ -158,7 +164,34 @@ def name(self):
...
+    def is_available(self) -> bool:
+        """Returns true if the device is currently available."""
+        return self._is_available_at_time(time.monotonic())
+
+    @property
+    def poll_interval(self) -> timedelta:
+        """Returns how often the device should be polled."""
+        if any(a.needs_polling for a in self.accessories):
+            return timedelta(minutes=1)
+        return timedelta(hours=24)
```
bdraco, line 177: The only device I have that we actually need to poll is the eve energy

aiohomekit/model/__init__.py
```diff
@@ -192,6 +206,19 @@ def hardware_revision(self) -> str:
...
+        """Check if there are any chars that need polling.
+
+        Currently this is only used for BLE devices that have
+        energy consumption characteristics.
+        """
+        for s in self.services:
+            for c in s.characteristics:
+                if c.type in NEEDS_POLLINGS_CHARS:
+                    return True
+        return False
```
bdraco, line 220: This needs a test

aiohomekit/controller/ble/connection.py
```diff
@@ -20,15 +20,20 @@
...
 
 logger = logging.getLogger(__name__)
 
-MAX_CONNECT_ATTEMPTS = 4
+MAX_CONNECT_ATTEMPTS = 3
+MAX_TRANSIENT_ERRORS = 8
+BLEAK_TIMEOUT = 9
+OVERALL_TIMEOUT = 10
+
+TRANSIENT_ERRORS = {"le-connection-abort-by-local", "br-connection-canceled"}
```
bdraco, line 36: Untyped exceptions 😢

Jc2k (approved):

PR #116, opened 2022-07-18
Do not reuse BleakClient per bleak authors recommendation

bdraco:
- Also fixes missed state changes


bdraco:
Looks like we can drop the double timeout now that we pass `BLEDevice` since it will never get into the discovery which can double the timeout

bdraco:
dare I say that its stable on linux now

bdraco:
I haven't had one failure to connect / put and I've been abusing it as much as possible for 20 minutes with these changes

Jc2k (approved):

bdraco:
Retest after letting it sit for a while looks good. Still reconnects 👍

bdraco:
Will push a release and bump hass in a few

PR #120, opened 2022-07-22
Allow passing in a bleak instance

bdraco:


Jc2k:
With zeroconf we ended up making it the callers responsibility to pass in a Zeroconf instance. Do we want to do that here for symmetry and not have a default?

bdraco:
> With zeroconf we ended up making it the callers responsibility to pass in a Zeroconf instance. Do we want to do that here for symmetry and not have a default?

I'm on the fence.  Maybe we change how this works once we have the COAP done?

bdraco:
Mostly on the fence because beta is Wed and I tend to get more conservative about making changes the closer we get to it.

Jc2k:
Haha fair point!

PR #121, opened 2022-07-24
Move all json/encode to hkjson, use orjson

bdraco:
I finally got enough hk devices on my network that I was
noticing the json overhead. Now that HASS supports orjson
we can use that instead


Jc2k (approved):
This looks great

PR #126, opened 2022-08-02
Fix hang at startup when bleak dbus connection stalls

bdraco:
Bumps bleak-retry-connector to >=1.1.1

Use async_timeout instead of asyncio.wait_for since
its much faster and we may end up retrying many times


bdraco:
I'm going to do a smaller one since Jc2k is asleep and I don't want to change any code without a review

PR #132, opened 2022-08-04
Add typing hints to accessories.serialize()

Jc2k:
This is the fix for one of only 2 failures left when turning on strict mypy for hkc's config_flow module.


bdraco (approved):

bdraco:
Probably needs an isort/black

PR #138, opened 2022-08-08
Handle BleakAbortedError in aiohomekit and rethrow as AccessoryDisconnectedError

bdraco:


Jc2k (approved):

bdraco:
I'm going to add a better error to `bleak-retry-connector` for this case https://community.home-assistant.io/t/2022-8-you-can-fix-it/446505/327?u=bdraco and rethrow as not found since its when the device disappears

bdraco:
I was really hoping some more test hardware would have shown up by now to look into the missing `format` but I guess thats going to have to wait until USPS finds the package.

PR #139, opened 2022-08-08
Drop mtu auto discovery

bdraco:
These turned out to be asymmetric and we will need
to wait for https://github.com/hbldh/bleak/issues/925
to be implmented to get the mtu with bluez.


Jc2k (approved):

bdraco:
Screen Shot 2022-08-08 at 9 17 44 AM

PR #141, opened 2022-08-09
Add support for bleak max_write_without_response_size and current bleak MTU

bdraco:
This is working with the current version of bleak as well and restores the performance that was lost in https://github.com/Jc2k/aiohomekit/pull/139

Also drops `.handle` calls since https://github.com/hbldh/bleak/issues/888

Needs https://github.com/hbldh/bleak/pull/930 for `max_write_without_response_size` but we can already get the MTU since the dict is accessible already.


Jc2k (approved):

bdraco:
Tested with 8 different devices. Pairing back to same speed 👍

Issue #135, opened 2022-08-07
Some services are missing when using BlueZ


bdraco:
`00000043-0000-1000-8000-0026BB765291` seen on MacOS but not Linux
[MACOS-homekit_controller-8629db23450ebc36b5589c3d6d653bd0-SYLVANIA Flex S-dbd7e25d780c89b5bf42b8736df88d14.json.txt](https://github.com/Jc2k/aiohomekit/files/9277594/MACOS-homekit_controller-8629db23450ebc36b5589c3d6d653bd0-SYLVANIA.Flex.S-dbd7e25d780c89b5bf42b8736df88d14.json.txt)
[LINUX-homekit_controller-8526992f05c02cbd3de691614408810f-SYLVANIA Flex S-9358632dace670c1fdbe50fe8c7dabba.json.txt](https://github.com/Jc2k/aiohomekit/files/9277595/LINUX-homekit_controller-8526992f05c02cbd3de691614408810f-SYLVANIA.Flex.S-9358632dace670c1fdbe50fe8c7dabba.json.txt)

bdraco:
https://github.com/WebBluetoothCG/web-bluetooth/issues/513

bdraco:
the service uuid is there, but its not being enumerated by bluez
```
[bluetooth]# info FA:8F:A4:3F:A3:2B
Device FA:8F:A4:3F:A3:2B (random)
	Name: SYLVANIA Flex SA C-3225
	Alias: SYLVANIA Flex SA C-3225
	Paired: no
	Trusted: no
	Blocked: no
	Connected: no
	LegacyPairing: no
	UUID: Vendor specific           (0000003e-0000-1000-8000-0026bb765291)
	UUID: Vendor specific           (00000043-0000-1000-8000-0026bb765291)
	UUID: Vendor specific           (00000055-0000-1000-8000-0026bb765291)
	UUID: Vendor specific           (000000a2-0000-1000-8000-0026bb765291)
	UUID: Generic Access Profile    (00001800-0000-1000-8000-00805f9b34fb)
	UUID: Generic Attribute Profile (00001801-0000-1000-8000-00805f9b34fb)
	UUID: Vendor specific           (1b7e8251-2877-41c3-b46e-cf057c562023)
	UUID: Vendor specific           (ae5d1e47-5c13-43a0-8635-82ad38a1381f)
	UUID: Vendor specific           (b0733e83-8434-4c00-a344-25d1c982a0ef)
[...]

PR #142, opened 2022-08-09
Improve error logging when a char or service is missing

bdraco:
Trying to work out https://github.com/Jc2k/aiohomekit/issues/135


bdraco:
I haven't had any luck with https://github.com/Jc2k/aiohomekit/issues/135 since the problem is deep in the bluez stack.  The improved logging will help us tell if other vendors are affected, but at this point it seems limited to just one vendor

Jc2k (approved):

PR #144, opened 2022-08-12
Implement bluetooth service caching

bdraco:
We now cache the services when reconnecting to reduce the time before we can start talking to the accessory.

The cache is never used when we need to fetch the GATT database or the ble device has changed
between connections.

The naïve address check has been replaced with the `ble_device_has_changed` helper which knows about
adapter changes.

### DIFF
--- a/aiohomekit/controller/ble/bleak.py
+++ b/aiohomekit/controller/ble/bleak.py
@@ -2,9 +2,10 @@ from __future__ import annotations
 
 import uuid
 
-from bleak import BleakClient, BleakError
+from bleak import BleakError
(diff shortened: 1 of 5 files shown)


bdraco:
Nanoleaf bulb seems a lot more reliable now.  No more failures to turn on because it timed out

bdraco:
We are effectively side stepping this issue https://github.com/hbldh/bleak/issues/927 as well since we only have to pay to to build that once now

Jc2k (approved):
Nice.

bdraco:
Thanks.

PR #150, opened 2022-08-13
Fix missing padding in SRP

bdraco:
Allows LIFX beam to pair

Refactoring to reduce int to bytes conversion to avoid the risk of random pairing failures when there are leading zeros in the key or salt since they would end up converted to short bytearrays

Screen Shot 2022-08-12 at 9 06 14 PM

### DIFF
--- a/aiohomekit/crypto/srp.py
+++ b/aiohomekit/crypto/srp.py
@@ -22,7 +22,6 @@ from __future__ import annotations
 import hashlib
 import math
 import os
-from sys import int_info
 from typing import Iterable
 
 # The K value for HK SRP is always the same because G and N are fixed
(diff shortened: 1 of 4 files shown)


Jc2k:
Oh nice!

bdraco:
Still need to clean it up some more and fix the server side refactoring but it works

bdraco:
I'm going to test with the box of gear from aliexpress in the morning as well

Jc2k (approved):

bdraco:
I still need to fix perform_pair_setup_part2 when we happen to generate a key that has a leading zero as that causes random pairing failures.

Jc2k:
Oh god 🤦

bdraco:
I think I could build a small script that generates keys for a few hours/minutes until it finds one that ends up with a leading zero with the public key. Then we can use that in a test

bdraco:
================================================================================================ 7000 passed in 1790.03s (0:29:50) =================================================================================================

PR #151, opened 2022-08-14
Update Zeroconf based backends to use _async_description_update

Jc2k:
This is still a draft, and mostly to help with conversations.

It changes ZeroconfController to populate a location cache of discoveries (`self.discoveries`) and keep it up to date with the latest parsed data.

`async_find` and `async_discover` then only operate on that cache.

And active pairings also get a `_async_description_update` callback (like bluetooth) with the latest `HomeKitService`.

So far the only benefits of this are:

 * The IP backend doesn't need to await to get the latest connection details as self.description will always be up to date
 * The CoAP backend doesn't rely on some magic coincidentally changing dict to pick up IP changes (and i think thats broken anyway).

But i also want to make some other changes, as below. 

# async_find changes (or new method)

Make a version of `async_find` that first yields the cache, then yields new hits from the firehose until a timeout.

This will fix the broken command line tools, and handle starting from a cold cache better in HA.

# Change pairing loading for Thread <-> BLE switching

Alter the way controllers load pairings so that we call this new `async_find` on every backend and return the first hit. In the case of Nanoleaf this will mean that we only need to reload the config entry to switch between BLE and Thread.

# Parity in char map and address handling between backends

Move some of the logic from BLE into share code (https://github.com/Jc2k/aiohomekit/blob/main/aiohomekit/controller/ble/pairing.py#L220-L268).

We should be able to drop reconnect_soon from HA and just use zeroconf direcly.

We should not need to explicitly do any char map stuff in HA, the pairing should update its char map automatically and notify HA that it needs to reprocess the char map and create new entitites.

### DIFF
--- a/aiohomekit/controller/coap/connection.py
+++ b/aiohomekit/controller/coap/connection.py
@@ -247,22 +247,12 @@ class CoAPHomeKitConnection:
         self.owner = owner
         self.pair_setup_client = None
 
-    async def reconnect_soon(self, updated_ip_port=None):
-        if updated_ip_port is None:
-            return
-        old_address = self.address
-        new_ip = updated_ip_port["AccessoryIP"]
-        new_port = updated_ip_port["AccessoryPort"]
-        new_address = f"[{new_ip}]:{new_port}"
-        logger.debug(
-            f"Device address update from zeroconf: old={old_address} new={new_address}"
-        )
-        if old_address != new_address:
-            self.address = new_address
-            await self.enc_ctx.coap_ctx.shutdown()
-            self.enc_ctx = None
-            # XXX can't .connect here w/o pairing_data
-        return
+    async def reconnect_soon(self):
+        description = self.owner.description
+        self.address = f"[{description.address}]:{description.port}"
+        await self.enc_ctx.coap_ctx.shutdown()
+        self.enc_ctx = None
+        # XXX can't .connect here w/o pairing_data
 
     async def do_identify(self):
         client = await Context.create_client_context()
--- a/aiohomekit/controller/coap/connection.py
+++ b/aiohomekit/controller/coap/connection.py
@@ -248,6 +248,8 @@ class CoAPHomeKitConnection:
         self.pair_setup_client = None
 
     async def reconnect_soon(self):
+        if not self.enc_ctx:
+            return
         description = self.owner.description
         self.address = f"[{description.address}]:{description.port}"
         await self.enc_ctx.coap_ctx.shutdown()
--- a/aiohomekit/controller/coap/pairing.py
+++ b/aiohomekit/controller/coap/pairing.py
@@ -51,10 +51,36 @@ class CoAPPairing(AbstractPairing):
         self.pairing_data = pairing_data
 
     def _async_description_update(self, description: Optional[HomeKitService]) -> None:
+        old_description = self.description
+
         super()._async_description_update(description)
-        async_create_task(
-            self.connection.reconnect_soon(f"{description.address}:{description.port}")
-        )
+
+        if not description:
+            return
+
+        endpoint_changed = False
+        if not old_description:
+            logger.debug("%s: Device rediscovered", self.id)
+            endpoint_changed = True
+        elif old_description.port != description.port:
+            logger.debug(
+                "%s: Device IP changed",
+                self.id,
+                old_description.address,
+                description.address,
+            )
+            endpoint_changed = True
+        elif old_description.address != description.address:
+            logger.debug(
+                "%s: Device port changed",
+                self.id,
+                old_description.port,
+                description.port,
+            )
+            endpoint_changed = True
+
+        if endpoint_changed:
+            async_create_task(self.connection.reconnect_soon())
 
     @property
     def is_connected(self):
--- a/aiohomekit/controller/coap/pairing.py
+++ b/aiohomekit/controller/coap/pairing.py
@@ -64,7 +64,7 @@ class CoAPPairing(AbstractPairing):
             endpoint_changed = True
         elif old_description.port != description.port:
             logger.debug(
-                "%s: Device IP changed",
+                "%s: Device IP changed from %s to %s",
                 self.id,
                 old_description.address,
                 description.address,
@@ -72,7 +72,7 @@ class CoAPPairing(AbstractPairing):
             endpoint_changed = True
         elif old_description.address != description.address:
             logger.debug(
-                "%s: Device port changed",
+                "%s: Device port changed from %s to %s",
                 self.id,
                 old_description.port,
                 description.port,
--- a/aiohomekit/controller/coap/pairing.py
+++ b/aiohomekit/controller/coap/pairing.py
@@ -17,7 +17,7 @@
 import asyncio
 from datetime import timedelta
 import logging
-from typing import Any
+from typing import Any, Optional
 
 from aiohomekit.controller.abstract import (
     AbstractController,
@@ -26,8 +26,10 @@ from aiohomekit.controller.abstract import (
 )
 from aiohomekit.exceptions import AccessoryDisconnectedError
 from aiohomekit.model import Accessories, AccessoriesState, Transport
+from aiohomekit.utils import async_create_task
 from aiohomekit.uuid import normalize_uuid
 
+from ...zeroconf import HomeKitService
 from .connection import CoAPHomeKitConnection
 
 logger = logging.getLogger(__name__)
@@ -47,6 +49,12 @@ class CoAPPairing(AbstractPairing):
         self.connection_future = None
         self.connection_lock = asyncio.Condition()
         self.pairing_data = pairing_data
+
+    def _async_description_update(self, description: Optional[HomeKitService]) -> None:
+        super()._async_description_update(description)
+        async_create_task(
+            self.connection.reconnect_soon(f"{description.address}:{description.port}")
+        )
 
     @property
     def is_connected(self):
@@ -219,12 +227,3 @@ class CoAPPairing(AbstractPairing):
     async def remove_pairing(self, pairingId):
         await self._ensure_connected()
         return await self.connection.remove_pairing(pairingId)
-
-    async def reconnect_soon(self) -> None:
-        # Conventiently Home Assistant mutates self.pairing_data for us
-        # So we can just re-read from pairing data.
-        # This is kinda gross but there is a longer term plan to refactor
-        # this API away, so will do for now
-        address = self.pairing_data["AccessoryAddress"]
-        port = self.pairing_data["AccessoryPort"]
-        await self.connection.reconnect_soon(f"{address}:{port}")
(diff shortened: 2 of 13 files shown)


aiohomekit/controller/coap/pairing.py
```diff
@@ -48,6 +50,12 @@ def __init__(
         self.connection_lock = asyncio.Condition()
         self.pairing_data = pairing_data
 
+    def _async_description_update(self, description: Optional[HomeKitService]) -> None:
+        super()._async_description_update(description)
+        async_create_task(
+            self.connection.reconnect_soon(f"{description.address}:{description.port}")
```
roysjosh, line 56: For CoAP at least the IPv6 address needs to be wrapped in square brackets, e.g. "[fe80::1234:2345:3456:4567]:5683". Also, the CoAP `reconnect_soon` needs fixing to accept a `str` vs a `dict`.

Jc2k: Ah cheers, i will definitely get a test env set up for this stuff soon.

Ideally we would make the IP/Port comparison here directly, without creating a task, and then only have to create a task if the IP had actually changed.

Jc2k: https://github.com/Jc2k/aiohomekit/pull/151/commits/133124b684f882595a6b6722b751af9241aefefc. Can probably do better still but don't want to touch CoAP too much before i've got thing working here :)

Jc2k:
`async_find` now search all backends in parallel, and yields the first discovery that we hit.

This fixes pairing on the command line.

aiohomekit/controller/coap/connection.py
```diff
@@ -247,22 +247,12 @@ def __init__(self, owner, host, port):
...
-        if old_address != new_address:
-            self.address = new_address
-            await self.enc_ctx.coap_ctx.shutdown()
-            self.enc_ctx = None
-            # XXX can't .connect here w/o pairing_data
-        return
+    async def reconnect_soon(self):
+        description = self.owner.description
+        self.address = f"[{description.address}]:{description.port}"
+        await self.enc_ctx.coap_ctx.shutdown()
```
bdraco, line 253: Is this going to be an issue when hass sees it https://github.com/home-assistant/core/blob/5f827f4ca64067ca886e0e4dcb0d1498689e1ded/homeassistant/components/homekit_controller/config_flow.py#L288?

aiohomekit/zeroconf.py
```diff
@@ -131,39 +144,64 @@ def __init__(
...
+        self,
+        zeroconf: Zeroconf,
+        service_type: str,
+        name: str,
+        state_change: ServiceStateChange,
+    ):
+        if service_type != self.hap_type:
+            return
+        info = AsyncServiceInfo(service_type, name)
+        async_create_task(self._async_handle_service(info))
```
bdraco, line 178: We could end up creating a lot of tasks here as we could get a `ServiceStateChange.Added` then a `ServiceStateChange.Updated`

Also need to check how `ServiceStateChange.Removed` is handled

bdraco:
Side note. I have a log full of `            logger.debug("Not a valid homekit device")` but I don't know what is causing them.  It would be nice to improve that message to have more detail

aiohomekit/zeroconf.py
bdraco: Maybe a 0.5s debouncer here that always runs again after the cooldown of _TIMEOUT_MS + 1s?

bdraco: I haven't found a good Debouncer lib.  Maybe we should pull the one out of HASS and publish it as a PyPi

Jc2k: Good point about Removed. Could either send a `None` description or ignore it. Ignore it easiest option to start with. Don't think we should do anything pro-active about it - if its just a bad network we might still have a somewhat stable TCP connection in flight.

Jc2k: Or... we won't update a pairing's description, but we'd probably remove it from `self.discoveries`

aiohomekit/controller/coap/connection.py
Jc2k: Nope, for 2 reasons:

* The HA PR to go with this will remove that call to reconnect_soon, as we'll just use zerocnf directly instead.
* This is a different reconnect_soon. CoAPPairing.reconnect_soon used to call this reconnect_soon, but not any more. The only reconnect_soon HA can call when this merges is an empty method, so even if i forget it won't hurt.

aiohomekit/zeroconf.py
Jc2k: I think i'm going to pull in the HA one into aiohomekit and get it working, then pull it out into a seperate package later. I'll burn all the time i have this evening setting up a new thing, and won't be hard to do later, just add a dep and change an import.

Jc2k: https://github.com/Jc2k/aiohomekit/pull/151/commits/3fc7783ee81ed5d89726801d204b8e0065466a66 pulls out the debouncer, and adds a async_trigger that runs in event loop but in sync context - reduces number of tasks created.

bdraco: Merged in main, doing live testing now

bdraco:
Basic testing show everything still works

Will hook up that problematic meross switch

bdraco:
I don't think its related but there is a race after the device is unpaired where it tries to still use the device

```
2022-08-14 22:44:51.445 DEBUG (MainThread) [aiohomekit.controller.ip.connection] 192.168.107.81: raw request: b'POST /pair-verify HTTP/1.1\r\nHost: 192.168.107.81\r\nContent-Length: 125\r\nContent-Type: application/pairing+tlv8\r\n\r\n\x06\x01\x03\x05x\x81#W\x82R\x8b3
2022-08-14 22:44:51.560 DEBUG (MainThread) [aiohomekit.controller.ip.connection] 192.168.107.81: raw response: bytearray(b'\x06\x01\x04\x07\x01\x02')
2022-08-14 22:44:51.560 DEBUG (MainThread) [aiohomekit.protocol.tlv] receiving [                                           
  6 (State): (1 bytes/<class 'bytearray'>) 0x04                                                                                
  7 (Error): (1 bytes/<class 'bytearray'>) 0x02 [Authentication]                                                             
]                                                                                                                               
                                                                                                                              
2022-08-14 22:44:51.560 ERROR (MainThread) [aiohomekit.utils] Failure running background task: Task-25963                      
Traceback (most recent call last):                                                                                           
  File "/usr/local/lib/python3.10/site-packages/aiohomekit/utils.py", line 28, in _handle_task_result                           
    task.result()                                                                                                             
  File "/usr/local/lib/python3.10/site-packages/aiohomekit/controller/ip/connection.py", line 533, in _reconnect                
    return await self._connect_once()                                                                                         
  File "/usr/local/lib/python3.10/site-packages/aiohomekit/controller/ip/connection.py", line 636, in _connect_once            
    request, expected = state_machine.send(response)                                                                         
  File "/usr/local/lib/python3.10/site-packages/aiohomekit/protocol/__init__.py", line 589, in get_session_keys                 
    handle_state_step(response_tlv, TLV.M4)                                                                                   
  File "/usr/local/lib/python3.10/site-packages/aiohomekit/protocol/__init__.py", line 87, in handle_state_step                 
    error_handler(tlv_dict[TLV.kTLVType_Error], f"step {expected_state}")                                                     
  File "/usr/local/lib/python3.10/site-packages/aiohomekit/protocol/__init__.py", line 61, in error_handler                     
    raise AuthenticationError(stage)                                                                                          
aiohomekit.exceptions.AuthenticationError: step bytearray(b'\x04')        
```

bdraco:
Seeing this on startup

```
2022-08-14 12:57:08.254 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved
Traceback (most recent call last):
  File "/Users/bdraco/home-assistant/venv/lib/python3.10/site-packages/aiohomekit/zeroconf.py", line 215, in async_find
    if discovery := await waiter:
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/bdraco/home-assistant/venv/lib/python3.10/site-packages/aiohomekit/zeroconf.py", line 214, in async_find
    async with async_timeout.timeout(timeout):
  File "/Users/bdraco/home-assistant/venv/lib/python3.10/site-packages/async_timeout/__init__.py", line 129, in __aexit__
    self._do_exit(exc_type)
  File "/Users/bdraco/home-assistant/venv/lib/python3.10/site-packages/async_timeout/__init__.py", line 212, in _do_exit
    raise asyncio.TimeoutError
asyncio.exceptions.TimeoutError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/bdraco/home-assistant/venv/lib/python3.10/site-packages/aiohomekit/zeroconf.py", line 218, in async_find
    raise AccessoryNotFoundError(
aiohomekit.exceptions.AccessoryNotFoundError: Accessory with device id 2b:5e:e9:f6:8e:b6 not found

```

bdraco:
```
./home-assistant.log.1:2022-08-14 12:29:07.080 DEBUG (MainThread) [aiohomekit.controller.ble.controller] Discovery for 2b:5e:e9:f6:8e:b6 already found
./home-assistant.log.1:aiohomekit.exceptions.AccessoryNotFoundError: Accessory with device id 2b:5e:e9:f6:8e:b6 not found
./home-assistant.log:2022-08-14 12:56:38.253 DEBUG (MainThread) [aiohomekit.controller.ble.controller] Discovery for 2b:5e:e9:f6:8e:b6 already found
./home-assistant.log:aiohomekit.exceptions.AccessoryNotFoundError: Accessory with device id 2b:5e:e9:f6:8e:b6 not found
```

bdraco:
I have no idea where '2b:5e:e9:f6:8e:b6' is coming from though as its not discoverable via zeroconf

bdraco:
Everything appears to be working.

That timeout exception keeps being throw periodically

Unpaired and paired a few things including a Meross 'OutLet'... thats new naming scheme :)

Screen Shot 2022-08-14 at 1 10 44 PM

bdraco:
So far so good with testing sans the timeout

aiohomekit/controller/controller.py
```diff
@@ -113,12 +114,28 @@ async def async_start(self) -> None:
...
+                done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
+                for result in done:
+                    try:
+                        return result.result()
+                    except AccessoryNotFoundError:
+                        continue
+        finally:
+            [task.cancel() for task in pending]
+            for task in pending:
+                try:
```
bdraco, line 135: Could use `with contextlib.suppress(asyncio.CancelledError)` 🤷‍♂️ but doesn't really matter

bdraco (approved):

aiohomekit/controller/controller.py
Jc2k: I was thinking of making a async_cancel(list[Task]) that did this because it's something I write a lot.

PR #153, opened 2022-08-15
Fix CoAP pairing

roysjosh:
A splat occurs during pair-setup when we attempt to unsubscribe from nothing.


aiohomekit/controller/coap/connection.py
```diff
@@ -625,6 +625,8 @@ def _unsubscribe_from_exit(
         return results
 
     async def unsubscribe_from(self, ids: list[tuple[int, int]]):
+        if len(ids) == 0:
```
bdraco, line 628:
```suggestion
        if not ids:
```

Jc2k (approved):

PR #154, opened 2022-08-15
Fix ble device not being passed from discovery to pairing when paired

bdraco:
We pass the connection on to the pairing once we finish pairing, however if the device is on the edge of range and it disconnects before the next ADV is received, device is unset and we cannot reconnect.

Also log the RSSI so we get a better idea why pairing is failing.

### DIFF
--- a/aiohomekit/controller/ble/discovery.py
+++ b/aiohomekit/controller/ble/discovery.py
@@ -69,11 +69,20 @@ class BleDiscovery(AbstractDiscovery):
         self._connection_lock = asyncio.Lock()
 
     @property
-    def name(self):
+    def name(self) -> str:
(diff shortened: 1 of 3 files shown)


aiohomekit/controller/ble/discovery.py
```diff
@@ -161,6 +174,7 @@ async def finish_pairing(pin: str) -> BlePairing:
             obj = self.controller.pairings[alias] = BlePairing(
                 self.controller,
                 pairing,
+                device=self.device,
```
bdraco, line 177: If we get disconnected before the next advertisement updates the device, everything blows up

bdraco:
Retested on linux as well just to be sure 👍

PR #156, opened 2022-08-16
Handle zeroconf c# changes directly not via notify_config_changed

Jc2k:
Gets rid of `notify_config_changed` entirely, prefering to do it internally based on triggers from zeroconf.

Note that BLE had `_process_config_changed` and `_async_process_config_changed`. One was trigger by bluetooth advertisements, one was triggered by HA. I have removed an arbritary one.

The thing I'm not yet decided on how to handle is state_num. It's important that its nestled in with the c# checks as that optimises the case where both have changed. But its a waste to create a task on IP and CoAP as they don't sue state_num. Could make it a sync callback, and only create a task in the BLE backend?

### DIFF
--- a/aiohomekit/controller/abstract.py
+++ b/aiohomekit/controller/abstract.py
@@ -65,13 +65,18 @@ class AbstractPairing(metaclass=ABCMeta):
     # and BLE advertisements), and also as AccessoryPairingID i pairing data.
     id: str
 
-    def __init__(self, controller: AbstractController) -> None:
+    def __init__(
+        self, controller: AbstractController, pairing_data: AbstractPairingData
+    ) -> None:
         self.controller = controller
         self.listeners: set[Callable[[dict], None]] = set()
         self.subscriptions: set[tuple[int, int]] = set()
         self.availability_listeners: set[Callable[[bool], None]] = set()
         self.config_changed_listeners: set[Callable[[int], None]] = set()
         self._accessories_state: AccessoriesState | None = None
+
+        self.id = pairing_data["AccessoryPairingID"]
+        self._load_accessories_from_cache()
 
     @property
     def accessories_state(self) -> AccessoriesState:
@@ -127,8 +132,8 @@ class AbstractPairing(metaclass=ABCMeta):
             )
 
         repopulate_accessories = False
-        if description and self.description:
-            if description.config_num > self.description.config_num:
+        if description:
+            if description.config_num > self.config_num:
                 logger.debug(
                     "%s: Config number has changed from %s to %s; char cache invalid",
                     self.id,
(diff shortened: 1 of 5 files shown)


aiohomekit/controller/abstract.py
```diff
@@ -109,11 +112,53 @@ def transport(self) -> Transport:
...
+            logger.debug(
+                "%s: Description updated: old=%s new=%s",
+                self.id,
+                self.description,
+                description,
+            )
+
+        repopulate_accessories = False
+        if description and self.description:
+            if description.config_num > self.description.config_num:
```
Jc2k, line 131: There is a case in the zeroconf transports where self.description starts as none. Those transports don't currently "fire" until we have seen one baseline descripton and then one change.

BLE solves this by making the description mandatory. We could do the same for zeroconf. So far we have not required zeroconf to be functional for "some" level of homekit support though (after pairing, anyway). The change would be relatively simple - just do a `async_find` into `load_pairing` and take the description from that.

Also need to compare against the stored config_num, but I want to decide what to do about ^ first.

bdraco: Considering how much trouble users have with transient MDNS problems (I maybe jaded from maintaining zeroconf), I would avoid making it mandatory and load from disk in this case.  With Ble we have to have the advertisement to connect to the device so we are stuck with the design there.

Jc2k: Damn i was hoping i was being jaded and you might give me a compelling reason why it wasn't a problem any more.

aiohomekit/controller/abstract.py
```diff
@@ -62,14 +65,19 @@ class AbstractPairing(metaclass=ABCMeta):
...
+    ) -> None:
         self.controller = controller
         self.listeners: set[Callable[[dict], None]] = set()
         self.subscriptions: set[tuple[int, int]] = set()
         self.availability_listeners: set[Callable[[bool], None]] = set()
         self.config_changed_listeners: set[Callable[[int], None]] = set()
         self._accessories_state: AccessoriesState | None = None
 
+        self.id = pairing_data["AccessoryPairingID"]
+        self._load_accessories_from_cache()
```
Jc2k, line 79: We now always load the cached accessories, if they are available.

aiohomekit/controller/abstract.py
```diff
@@ -109,11 +117,56 @@ def transport(self) -> Transport:
...
+            logger.debug(
+                "%s: Description updated: old=%s new=%s",
+                self.id,
+                self.description,
+                description,
+            )
+
+        repopulate_accessories = False
+        if description:
+            if description.config_num > self.config_num:
```
Jc2k, line 136: And then we can compare to self.config_num instead of the previous description, which fixes the problem I had with the previous patch.

bdraco (approved):
Confirmed the c# change is seen

https://hastebin.com/jutigosobi.swift

PR #166, opened 2022-08-25
Don't ignore Thread Transport which skips Thread Status

roysjosh:


bdraco (approved):
I added this originally to reduce the sync time, but we are actually using it now so good to change 👍

PR #173, opened 2022-09-11
Bump bleak-retry-connector to implement a smart backoff 

bdraco:
When the device is still connected for a few ms after an abort we need to wait for it to fully disconnect

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -777,7 +777,7 @@ ifaddr = ">=0.1.7"
 [metadata]
 lock-version = "1.1"
 python-versions = "^3.9"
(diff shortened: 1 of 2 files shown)


Jc2k (commented):
Should this be 1.14 or is the last one wrong?

bdraco:
Should be  1.14 .. good catch.  I just did 8 of these and at least I only fat fingered this one :)

Jc2k (approved):

bdraco:
Guess I spoke too soon.  Looks like I coped that to yalexs-ble as well. Thanks for catching that 🚀

PR #175, opened 2022-09-11
Leave connection open if ble device changes

### LINKED ISSUE #174: Don't close the connection when the ble device has changed if already connected
I was doing some testing with adapters very close to each other. Hopefully this doesn't happen in the real world, but we all know it will.

If we already have a connection open, and the ble device changes, closing it doesn't seem make much sense since we already know we have a working connection to the device.

https://github.com/Jc2k/aiohomekit/blob/d6e6a58b6755c1ed46d764f7264941bcd6c5c56d/aiohomekit/controller/ble/pairing.py#L236

I preserved this original logic but I think we should rip it out, but I don't know why it was added originally.

bdraco:
- We will reconnect with the new one the next time, however if the device is already connected there is no reason to switch until the next time.

Fixes #174

### DIFF
--- a/aiohomekit/controller/ble/pairing.py
+++ b/aiohomekit/controller/ble/pairing.py
@@ -233,7 +233,7 @@ class BlePairing(AbstractPairing):
             original_path = original_device.details["path"]
             new_path = device.details["path"]
         logger.debug(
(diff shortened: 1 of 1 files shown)


Jc2k (approved):

bdraco:
This made the test nanoleaf bulb a lot more stable 👍

PR #193, opened 2022-10-16
Switch to using retry_bluetooth_connection_error from bleak-retry-connector

bdraco:
Shifts the maint burden to the lib which already has tests for this


Jc2k (approved):

bdraco:
Thanks. I'll get this released and hass bumped after I land and that should get us fully ready for bleak 0.19

PR #201, opened 2022-10-22
fix: skip services that are missing an iid

bdraco:
If we create them they will get assigned the next available one which will conflict with the actual one. We need to ignore these services as they are not homekit services but services defined by the vendor


Jc2k (approved):

bdraco:
Everything looks solid now.

Going to push

PR #202, opened 2022-10-23
Allow specifying the number of attempts with async_populate_accessories_state

bdraco:
Currently we only use this value for BLE connections

The goal is to fail quickly at startup so we do not block HASS startup and let the retry logic take over after startup to get the device up and going


Jc2k (approved):

bdraco:
This is working well in testing 👍

bdraco:
Will push a tag shortly. Just doing some final testing

PR #203, opened 2022-10-23
Fallback to disconnected events when receiving encrypted notify without having the key yet

bdraco:
If the device is asleep at startup we will not have the key when it wakes up and sends us a notification.


Jc2k (approved):

bdraco:
Will do a push for this in a minute.  Getting to be a long day. 💤

PR #204, opened 2022-10-24
Read global state number when updating chars

### LINKED ISSUE #199: When connected and after gatt notify update read and update the global state num
We can probably reduce the number of disconnected events since if the adv would have been updated while we are connected we don't know the new global state number, and then we disconnect, see the adv and connect again

bdraco:
closes #199

### DIFF
--- a/aiohomekit/controller/ble/pairing.py
+++ b/aiohomekit/controller/ble/pairing.py
@@ -125,6 +125,10 @@ GENERATE_BROADCAST_KEY_PAYLOAD = (
     + b"\x00"
 )
 
+GET_ALL_PARAMS_PAYLOAD = (
+    bytes([HAP_BLE_PROTOCOL_CONFIGURATION_REQUEST_TLV.GetAllParams]) + b"\x00"
+)
+
 WrapFuncType = TypeVar("WrapFuncType", bound=Callable[..., Any])
 
 
@@ -292,10 +296,14 @@ class BlePairing(AbstractPairing):
         super()._async_description_update(description)
 
     async def _async_request(
-        self, opcode: OpCode, char: Characteristic, data: bytes | None = None
+        self,
+        opcode: OpCode,
+        char: Characteristic,
+        data: bytes | None = None,
+        iid: int | None = None,
     ) -> bytes:
         async with self._ble_request_lock:
-            return await self._async_request_under_lock(opcode, char, data)
+            return await self._async_request_under_lock(opcode, char, data, iid)
 
     async def _async_request_under_lock(
         self,
@@ -451,6 +459,8 @@ class BlePairing(AbstractPairing):
                 self.rssi,
             )
             try:
+                await self.get_global_state_number()
+
                 results = await self.get_characteristics(list(self.subscriptions))
             except (
                 AccessoryDisconnectedError,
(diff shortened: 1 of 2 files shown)


bdraco:
Well the key is in there as well so we don't need to calculate it!

2022-10-23 21:52:35.690 WARNING (MainThread) [aiohomekit.controller.ble.pairing] VOCOlinc-VS1-185620 [E5:AB:B3:23:65:23] (id=E4:2F:25:E5:71:C5): Got global state number response: {1: bytearray(b'\xaf\x01'), 2: bytearray(b'\x01'), 3: bytearray(b'\xe4/%\xe5q\xc5'), 4: bytearray(b'\xee\x8c^5\x9d\xa0\x17,,i\xe3\xdb\x90\x83\x93LXI{v\xb1\xbf8ha\xac\xd1o)|,\xf4')}

bdraco:
This helps a lot at startup since we don't trigger a catch up poll right after the first disconnect since the state num will be correct instead of only updated from the adv

bdraco:
What would happen before would be we would connect, sync state, not update the global state number, disconnect, and then the accessory advertises the state right away after disconnect so we would connect again and sync again

Now that we get the number during the connection we don't connect twice

aiohomekit/controller/ble/pairing.py
```diff
@@ -125,9 +127,31 @@
...
+@dataclass
+class ProtocolParams:
+
+    state_number: int
+    config_number: int
+    advertising_id: bytes
+    broadcast_key: bytes
+
+
+class ProtocolParamsTLV(IntEnum):
```
bdraco, line 146: Need to move this

aiohomekit/controller/ble/pairing.py
```diff
@@ -125,9 +127,31 @@
...
 
+GET_ALL_PARAMS_PAYLOAD = (
+    bytes([HAP_BLE_PROTOCOL_CONFIGURATION_REQUEST_TLV.GetAllParams]) + b"\x00"
+)
+
 WrapFuncType = TypeVar("WrapFuncType", bound=Callable[..., Any])
 
 
+@dataclass
+class ProtocolParams:
```
bdraco, line 138: Probably move this as well but may be ok here ?

aiohomekit/controller/ble/pairing.py
```diff
@@ -714,6 +753,51 @@ async def _populate_char_values(self, config_changed: bool) -> None:
...
+            state_number=int.from_bytes(
+                response[ProtocolParamsTLV.GLOBAL_STATE_NUMBER], "little"
+            ),
+            config_number=int.from_bytes(
+                response[ProtocolParamsTLV.CONFIGURATION_NUMBER], "little"
+            ),
+            advertising_id=response[ProtocolParamsTLV.ADVERTISING_ID],
+            broadcast_key=response[ProtocolParamsTLV.BROADCAST_KEY],
+        )
+        logger.warning("%s: Got protocol params: %s", self.name, protocol_params)
```
bdraco, line 798: Downgrade to debug

bdraco:
Too exhausted to finish this tonight. Will try to pick it up tomorrow if Monday isn't too busy

Jc2k (approved):

bdraco:
Tested is solid on the esp32 and local.

Will push a tag shortly

PR #209, opened 2022-10-24
Add workaround for older eve devices

bdraco:
If this turns out to be more common we will probably need a function to handle additional workarounds. Since iOS is forgiving for this we should be as well

fixes https://github.com/home-assistant/core/issues/80177

### DIFF
--- a/aiohomekit/controller/ble/pairing.py
+++ b/aiohomekit/controller/ble/pairing.py
@@ -644,8 +644,14 @@ class BlePairing(AbstractPairing):
                     continue
 
                 decoded = CharacteristicTLV.decode(signature).to_dict()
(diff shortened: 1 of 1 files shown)


Jc2k (approved):

bdraco:
something else might be going on here

```
2022-10-24 10:17:47.110 DEBUG (MainThread) [aiohomekit.controller.ble.pairing] Eve Room 0881 [EB:4A:D4:93:68:EF] (id=6C:D6:F8:C8:96:56): char: 00000014-0000-1000-8000-0026bb765291 (Handle: 16481): Hardcopy Data Channel decoded: {'type': '140000100080000026BB765291', 'iid': None, 'perms': ['pw'], 'format': 'bool'}
```

PR #214, opened 2022-10-27
Fix for various races with encrypted notifications

bdraco:
Also fixes a race where was_connected was wrong because we were waiting for the lock and by the time is was acquired the value was stale and we had to reconnect.

Before these changes I was able to get the eve button to fallback to disconnected events.  Now I can mash it over and over and I get broadcast events for everything

Reduces the latency of disconnected events by calling the listeners after each char is read to avoid waiting for everything to be read

Ensure we re-read the GSN after starting notify and after a notify so we are always in sync

### DIFF
--- a/aiohomekit/controller/ble/pairing.py
+++ b/aiohomekit/controller/ble/pairing.py
@@ -476,8 +476,7 @@ class BlePairing(AbstractPairing):
                 self.rssi,
             )
             try:
-                protocol_param = await self.get_all_protocol_params()
-                results = await self.get_characteristics(list(self.subscriptions))
+                protocol_param, results = self._process_disconnected_events_with_retry()
             except (
                 AccessoryDisconnectedError,
                 *BLEAK_EXCEPTIONS,
@@ -496,6 +495,19 @@ class BlePairing(AbstractPairing):
 
             if protocol_param:
                 self.description.state_num = protocol_param.state_number
+
+    @operation_lock
+    @retry_bluetooth_connection_error()
+    @restore_connection_and_resume
+    async def _process_disconnected_events_with_retry(
+        self,
+    ) -> tuple[ProtocolParams | None, dict[tuple[int, int], dict[str, Any]]]:
+        accessory_chars = self.accessories.aid(1).characteristics
+        protocol_param = await self._get_all_protocol_params()
+        results = await self._get_characteristics_while_connected(
+            [accessory_chars.iid(iid) for _, iid in self.subscriptions]
+        )
+        return protocol_param, results
 
     def _async_notification(self, data: HomeKitEncryptedNotification) -> None:
         """Receive a notification from the accessory."""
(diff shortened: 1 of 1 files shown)


aiohomekit/controller/ble/pairing.py
```diff
@@ -467,8 +476,10 @@ async def _process_disconnected_events(self) -> None:
                 self.rssi,
             )
             try:
-                protocol_param = await self.get_all_protocol_params()
-                results = await self.get_characteristics(list(self.subscriptions))
+                (
+                    protocol_param,
+                    results,
+                ) = await self._process_disconnected_events_with_retry()
```
bdraco, line 482: I split this out so it was only wrapped once instead of two functions with their own wrappers so we don't retry twice, and then if it does raise we are still handling the exception

aiohomekit/controller/ble/pairing.py
```diff
@@ -576,28 +600,39 @@ def _async_get_service_signature_char(self) -> Characteristic | None:
...
-            service_iid,
-        )
-        try:
-            await self._async_request_under_lock(
-                OpCode.PROTOCOL_CONFIG,
-                hap_char,
-                GENERATE_BROADCAST_KEY_PAYLOAD,
-                iid=service_iid,
+        logger.debug("%s: Setting broadcast encryption key", self.name)
+        if self._ble_request_lock.locked():
```
bdraco, line 604: We always needed a lock when we called this so I moved it inside the function

aiohomekit/controller/ble/pairing.py
```diff
@@ -851,17 +879,17 @@ async def _populate_accessories_and_characteristics(
             if self._shutdown:
                 return
 
-            was_connected = self.client and self.client.is_connected
-            self._restore_pending |= not was_connected
             self._tried_to_connect_once = True
+
+            made_connection = await self._ensure_connected(attempts)
```
bdraco, line 884: I made ensure_connected return if it had to make a connection or not since there was a race here since we didn't hold the lock while we checked for `was_connected` above

aiohomekit/controller/ble/pairing.py
```diff
@@ -944,19 +972,12 @@ async def _async_restore_subscriptions(self) -> None:
...
             return
 
-        logger.debug("%s: Setting broadcast encryption key", self.name)
-        async with self._ble_request_lock:
-            await self._async_set_broadcast_encryption_key()
-        long_term_pub_key_hex: str = self.pairing_data["iOSDeviceLTPK"]
-        long_term_pub_key_bytes = bytes.fromhex(long_term_pub_key_hex)
-        self._broadcast_decryption_key = BroadcastDecryptionKey(
-            self._derive(long_term_pub_key_bytes, b"Broadcast-Encryption-Key")
-        )
```
bdraco, line 954: Moved this into `_async_set_broadcast_encryption_key` so we are sure its never out of sync

Jc2k (approved):
Looks good 👍

bdraco:
Retest after sleep looks good

Will push a tag and bump ha shortly

bdraco:
Release failed because PyPI uploads are down

https://status.python.org/incidents/kc846w6htvys

PR #218, opened 2022-10-29
Restore broadcast_key between restarts

### LINKED ISSUE #215: Save the last broadcast key between runs
The wemo stage seems to not accept connections as frequently anymore once broadcast events have been subscribed to

bdraco:
TLDR: This fixes battery powered BLE devices that sleep for long intervals

Some accessories will sleep for a long time and only send broadcasted events which makes them have very long connection intervals to save battery. Since we need to connect to get a new broadcast key we now save the broadcast key between restarts to ensure we can decrypt the advertisements coming in even though we cannot make a connection to the device during startup. When we get a disconnected event later we will try again to connect and the device will be awake which will trigger a full sync

Additional details: https://github.com/home-assistant/core/pull/81211

fixes #215

### DIFF
--- a/aiohomekit/controller/ble/pairing.py
+++ b/aiohomekit/controller/ble/pairing.py
@@ -85,7 +85,7 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-# The discover timeout is how longer we will wait for and advertisement to be
+# The discover timeout is how long we will wait for and advertisement to be
 # received. If we don't get it in this time we will try again later since
 # the scanner is always running anyways.
 DISCOVER_TIMEOUT = 10
--- a/aiohomekit/controller/ble/pairing.py
+++ b/aiohomekit/controller/ble/pairing.py
@@ -277,13 +277,6 @@ class BlePairing(AbstractPairing):
             return timedelta(minutes=5)
         return timedelta(hours=24)
 
-    @property
-    def broadcast_key(self) -> bytes | None:
-        """Return the broadcast key."""
-        if not self._accessories_state:
-            return None
-        return self._accessories_state.broadcast_key
-
     def _is_available_at_time(self, monotonic: float) -> bool:
         """Check if we are considered available at the given time."""
         return self.is_connected or monotonic - self._last_seen < AVAILABILITY_INTERVAL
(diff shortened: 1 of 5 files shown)


aiohomekit/controller/ble/pairing.py
```diff
@@ -636,9 +647,13 @@ async def _async_set_broadcast_encryption_key(self) -> None:
...
         long_term_pub_key_bytes = bytes.fromhex(long_term_pub_key_hex)
-        self._broadcast_decryption_key = BroadcastDecryptionKey(
-            self._derive(long_term_pub_key_bytes, b"Broadcast-Encryption-Key")
+        broadcast_key_bytes = self._derive(
+            long_term_pub_key_bytes, b"Broadcast-Encryption-Key"
         )
+        self._broadcast_decryption_key = BroadcastDecryptionKey(broadcast_key_bytes)
+        if self._accessories_state and self.broadcast_key != broadcast_key_bytes:
+            self._accessories_state.broadcast_key = broadcast_key_bytes
+            self._callback_and_save_config_changed(self.config_num)
```
bdraco, line 656: todo: check if we need this

bdraco: we do

aiohomekit/controller/ble/pairing.py
```diff
@@ -85,7 +85,10 @@
 
 logger = logging.getLogger(__name__)
 
-DISCOVER_TIMEOUT = 30
+# The discover timeout is how longer we will wait for and advertisement to be
+# received. If we don't get it in this time we will try again later since
+# the scanner is always running anyways.
```
bdraco, line 90: Not much point in waiting for something that may never come if the device is sleeping

Jc2k (approved):

aiohomekit/controller/ble/pairing.py
```diff
@@ -85,7 +85,10 @@
 
 logger = logging.getLogger(__name__)
 
-DISCOVER_TIMEOUT = 30
+# The discover timeout is how longer we will wait for and advertisement to be
```
bdraco, line 88:
```suggestion
# The discover timeout is how long we will wait for and advertisement to be
```

bdraco:
Still need to retest with IP accessories. I don't expect it will matter but good to check

bdraco:
All good 👍

bdraco:
Will push a tag, retest, and update the HA PR shortly

PR #222, opened 2022-10-30
Only update characteristics that support disconnected events when processing disconnected events

### LINKED ISSUE #221: _process_disconnected_events_with_retry should skip chars that do not support disconnected events

bdraco:
We didn't have a way to tell before, but now that we have these values we can use them.

fixes #221


Jc2k (approved):

bdraco:
Thanks.  Working well overnight

PR #227, opened 2022-10-31
Ignore link local addresses

Jc2k:
This should fix https://github.com/home-assistant/core/issues/80211 by replicating the old HA behaviour of dropping link local addresses.


bdraco:
Looks good.  I'll test later today

bdraco (approved):
Verified link local ignored as expected

PR #230, opened 2022-11-01
Fix missing operation lock for BLE notifications

bdraco:


aiohomekit/controller/ble/pairing.py
```diff
@@ -638,6 +640,7 @@ def _async_get_service_signature_char(self) -> Characteristic | None:
 
     async def _async_set_broadcast_encryption_key(self) -> None:
         """Get the broadcast key for the accessory."""
+        assert self._operation_lock.locked(), "Should be called with operation lock"
```
bdraco, line 643:
```suggestion
        assert self._operation_lock.locked(), "_operation_lock should be locked"
```

Jc2k (approved):

aiohomekit/controller/ble/pairing.py
```diff
@@ -1055,6 +1059,9 @@ async def _async_restore_subscriptions(self) -> None:
             self._restore_pending = False
             return
 
+        assert (
+            self._operation_lock.locked()
+        ), "Should be called under the operation lock"
```
bdraco, line 1064:
```suggestion
        ), "_operation_lock should be locked"
```

bdraco:
Noticed this since the subscribe to gatt notify causing a disconnect in the middle of a competing op that shouldn't have been competing 😨

bdraco:
10 restarts.  No more random failures due to gatt notify in the middle of a sync 👍

bdraco:
Pushing a tag.  Last one before I have to go to the airport and fly to europe

PR #231, opened 2022-11-01
Notify listeners of new state on successful BLE write

bdraco:
It can take 3-10s depending on how slow the ble connection is for HA to update after a successful write. This sometimes makes the switch turn back off in HA after it was successful so we should callback the change as soon as we know the OpStatus is successful to avoid the flip-flop.  With the nanoleaf devices this can even take > 10s because of how many chars need to be read to sync.

This change ensures HA more accurately reflects the state of the device since the light strip would light up and than HA changed state 10-15s later

### DIFF
--- a/aiohomekit/controller/ble/pairing.py
+++ b/aiohomekit/controller/ble/pairing.py
@@ -1334,7 +1334,7 @@ class BlePairing(AbstractPairing):
                     }
                 if single_result["status"] == HapStatusCode.SUCCESS:
                     for listener in self.listeners:
-                        listener(single_result)
+                        listener({result_key: single_result})
                 results[result_key] = single_result
 
         return results
--- a/aiohomekit/controller/ble/pairing.py
+++ b/aiohomekit/controller/ble/pairing.py
@@ -1288,6 +1288,7 @@ class BlePairing(AbstractPairing):
 
             for aid, iid, value in characteristics:
                 char = accessory_chars.iid(iid)
+                result_key = (aid, iid)
                 logger.debug(
                     "%s: Writing characteristics: iid=%s value=%s",
                     self.name,
@@ -1309,7 +1310,11 @@ class BlePairing(AbstractPairing):
                         OpCode.CHAR_TIMED_WRITE, char, payload
                     )
                     await self._async_request_under_lock(OpCode.CHAR_EXEC_WRITE, char)
-
+                    single_result = {
+                        "status": HapStatusCode.SUCCESS,
+                        "description": HapStatusCode.SUCCESS.description,
+                        "value": value,
+                    }
                 elif CharacteristicPermissions.paired_write in char.perms:
                     payload = TLV.encode_list(
                         [(HAP_TLV.kTLVHAPParamValue, to_bytes(char, value))]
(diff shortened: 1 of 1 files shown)


bdraco:
I see it writing but not changing quick enough

```
Nov 01 22:47:52 homeassistant homeassistant[457]: 2022-11-01 17:47:52.292 DEBUG (MainThread) [aiohomekit.controller.ble.pairing] Nanoleaf Strip 2D2S [CF:DF:A5:C4:F4:AC] (id=EA:B2:45:87:02:49): Writing characteristics: [(1, 51, 0)]; rssi=-59
Nov 01 22:47:52 homeassistant homeassistant[457]: 2022-11-01 17:47:52.292 DEBUG (MainThread) [aiohomekit.controller.ble.pairing] Nanoleaf Strip 2D2S [CF:DF:A5:C4:F4:AC] (id=EA:B2:45:87:02:49): Writing characteristics: iid=51 value=0
Nov 01 22:47:52 homeassistant homeassistant[457]: 2022-11-01 17:47:52.292 DEBUG (MainThread) [aiohomekit.protocol.tlv] sending [
Nov 01 22:47:52 homeassistant homeassistant[457]:   1 (Identifier): (1 bytes/<class 'bytes'>) b'\x00'
Nov 01 22:47:52 homeassistant homeassistant[457]: ]
Nov 01 22:47:52 homeassistant homeassistant[457]: 
Nov 01 22:47:52 homeassistant homeassistant[457]: 2022-11-01 17:47:52.292 DEBUG (MainThread) [aiohomekit.controller.ble.client] Queuing fragment for write: b'\x00\x02\xb93\x00\x03\x00\x01\x01\x00'

```

Out of time to finish this one though since I'm about to go to the airport.

Will pack some ble devices.


draft for now

bdraco:
Might be able to finish it before takeoff with ssh tunnels :)

bdraco:
Light control is so much more responsive now

Jc2k (approved):

Jc2k (commented):
I thought we already automatically updated the state on the HA side, as there was a similar delay in IP events (you don't get events for changes you made, so without that logic you don't see the change till the next poll). It's in the put_characteristic helper on the Hkdevice if I remember correctly.

So I guess ideally we'd remove the logic from the Ha side and make this change to the coap and ip backends as well.

bdraco:
I’m thinking the best path for this is to get this in 2022.11.x since the delay is very noticeable on ble since the protocol latency is at least an order of magnitude worse

Then start a new PR for 2022.12.x to adjust all the backends and HA since I think it’s too risky to make HA changes that remove logic in a patch release without some beta runway.

Jc2k:
Sounds good 👍

bdraco:
Also I don’t have a reliable CoAP setup so will probably need some help with validating that change.

IP and BLE obviously won’t be an issue to test 😉

bdraco:
Will pick this back up after any release issues

Jc2k:
Cheers. :+1: have 6 thread devices online, 7 when I setup the Wemo, so happy to test. Will test some BLE and IP stuff here as well anyway.

PR #235, opened 2022-11-03
Fallback to disconnected events if we cannot decrypt the notification

### LINKED ISSUE #234: Process notifications as disconnected events if we cannot decrypt so we don't get out of sync
logger.warning(
            "%s: Received notification but could not decrypt: %s", self.name, data
        )

bdraco:
fixes #234


Jc2k (approved):

bdraco:
I don't know why I didn't think of this in the first place 🤷

bdraco:
thanks

PR #237, opened 2022-11-03
Callback listeners on put for all controllers

### LINKED ISSUE #236: Modify COAP and IP backends to send state right away like we did for BLE

bdraco:
closes #236

### DIFF
--- a/aiohomekit/controller/coap/pairing.py
+++ b/aiohomekit/controller/coap/pairing.py
@@ -19,6 +19,7 @@ import asyncio
 from datetime import timedelta
 import logging
 from typing import Any, Iterable
+from aiohomekit.model.characteristics import CharacteristicPermissions
 
 from aiohomekit.controller.abstract import AbstractController, AbstractPairingData
 from aiohomekit.exceptions import AccessoryDisconnectedError
@@ -199,7 +200,12 @@ class CoAPPairing(ZeroconfPairing):
         listener_update = {tuple[int, int]: dict[str, Any]}
         for characteristic in characteristics:
             aid, iid, value = characteristic
-            if response_status.get((aid, iid), 0) == 0:
+            accessory_chars = self.accessories.aid(aid).characteristics
+            char = accessory_chars.iid(iid)
+            if (
+                response_status.get((aid, iid), 0) == 0
+                and CharacteristicPermissions.paired_read in char.perms
+            ):
                 listener_update[(aid, iid)] = {"value": value}
 
         if listener_update:
--- a/aiohomekit/controller/coap/pairing.py
+++ b/aiohomekit/controller/coap/pairing.py
@@ -19,11 +19,12 @@ import asyncio
 from datetime import timedelta
 import logging
 from typing import Any, Iterable
-from aiohomekit.model.characteristics import CharacteristicPermissions
-from aiohomekit.protocol.statuscodes import HapStatusCode
+
 from aiohomekit.controller.abstract import AbstractController, AbstractPairingData
 from aiohomekit.exceptions import AccessoryDisconnectedError
 from aiohomekit.model import Accessories, AccessoriesState, Transport
+from aiohomekit.model.characteristics import CharacteristicPermissions
+from aiohomekit.protocol.statuscodes import HapStatusCode
 from aiohomekit.utils import async_create_task
 from aiohomekit.uuid import normalize_uuid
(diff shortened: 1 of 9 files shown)


aiohomekit/controller/ip/pairing.py
```diff
@@ -280,28 +275,33 @@ async def put_characteristics(self, characteristics):
...
-            iid = characteristic[1]
-            value = characteristic[2]
-            characteristics_set.add(f"{aid}.{iid}")
-            data.append({"aid": aid, "iid": iid, "value": value})
-        data = {"characteristics": data}
-
-        response = await self.connection.put_json("/characteristics", data)
+            aid, iid, value = characteristic
+            char_payload.append({"aid": aid, "iid": iid, "value": value})
+            listener_update[(aid, iid)] = {"value": value}
```
bdraco, line 283: We should only add chars that we can read back

aiohomekit/controller/coap/pairing.py
```diff
@@ -198,9 +190,22 @@ async def get_characteristics(
...
+    ) -> dict[tuple[int, int], dict[str, Any]]:
         await self._ensure_connected()
-        return await self.connection.write_characteristics(characteristics)
+        response_status = await self.connection.write_characteristics(characteristics)
+
+        listener_update = {tuple[int, int]: dict[str, Any]}
+        for characteristic in characteristics:
+            aid, iid, value = characteristic
+            if response_status.get((aid, iid), 0) == 0:
+                listener_update[(aid, iid)] = {"value": value}
```
bdraco, line 203: We should only add chars that we can read back

bdraco:
I just realized we are polling chars that send events as well.

We should probably only poll for chars that don't have the ev flag once we do the initial sync

Jc2k:
> I just realized we are polling chars that send events as well.
> 
> 
> 
> We should probably only poll for chars that don't have the ev flag once we do the initial sync

I didn't do this originally until I had more confidence in the event code.

Now the biggest fear are stale TCP streams - polling has been an ok way of finding out if the other side crashes without doing an TCP RST on their side or if there was some other network failure (or they just got unplugged).

We could do a TCP keep alive but that's not part of spec so I don't know if it's widely support or any better than polling.

I think IP does have some version of the GSN, which made me wonder if it had zeroconf based disconnected events, but I'm not sure I saw it in the spec, and I'm not sure all my devices have the field in their zeroconf data.

Jc2k:
Can't remember what we do atm, but for CoAP, it's UDP so no stream so I'm less worried about a stalled connection though still a bit worried about how we detect the device is unavailable

Would be better not to poll if we do because thread has battery powered devices.

Jc2k:
Maybe the zeroconf recording dropping drives unavailability?

bdraco:
> Maybe the zeroconf recording dropping drives unavailability?

That would make a lot of sense. If the zeroconf PTR expires we could ask the device if is still there and than mark it unavailable.

bdraco:
Split the zeroconf / polling issue off into https://github.com/Jc2k/aiohomekit/issues/243

aiohomekit/controller/ip/pairing.py
```diff
@@ -280,28 +276,36 @@ async def put_characteristics(self, characteristics):
...
-            iid = characteristic[1]
-            value = characteristic[2]
-            characteristics_set.add(f"{aid}.{iid}")
-            data.append({"aid": aid, "iid": iid, "value": value})
-        data = {"characteristics": data}
-
-        response = await self.connection.put_json("/characteristics", data)
+            aid, iid, value = characteristic
+            char_payload.append({"aid": aid, "iid": iid, "value": value})
+            accessory_chars = self.accessories.aid(aid).characteristics
```
bdraco, line 284: This does a linear search under the hood which could be an issue with many acessories.  We should optimize that

bdraco: https://github.com/Jc2k/aiohomekit/pull/248

bdraco:
The tests that use the AccessoryServer kept failing so I went ahead and changed them to use the next available port instead and they seem to be a lot more reliable and faster now

bdraco:
I've tested this with IP and BLE but I don't have COAP so I can't test there.

Jc2k (approved):

Jc2k:
Working here with a CoAP Eve Thermo

PR #241, opened 2022-11-06
Improve BLE pairing reliability, especially with esp32 proxies

bdraco:
Overrides the default implementation of get_primary_name
to get name from the advertisement if the
accessories are not available which happens
frequently with the ESPHome proxies because
they cannot currently connect right away again
after disconnecting.

We want to avoid raising there if possible
since this is called right after pairing
and if we raise the pairing will fail but
the device will still be paired and the user
will have to un-pair it manually with a factory
reset.

Also modifies the handle check to
handle no services being available yet.

### DIFF
--- a/aiohomekit/controller/ble/pairing.py
+++ b/aiohomekit/controller/ble/pairing.py
@@ -882,20 +882,12 @@ class BlePairing(AbstractPairing):
         will have to un-pair it manually with a factory
         reset.
         """
-        if not self.accessories:
-            if self.description:
-                self._accessories_state = AccessoriesState(
-                    Accessories(), -1, self.broadcast_key
-                )
-                return self.description.name
-            parsed = await self.list_accessories_and_characteristics()
-        else:
-            parsed = self.accessories
-
-        accessory_info = parsed.aid(1).services.first(
-            service_type=ServicesTypes.ACCESSORY_INFORMATION
-        )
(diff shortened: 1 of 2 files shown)


bdraco:
I can get the Nanoleaf bulbs to pair with this change using the ESP32 a

Jc2k (approved):

aiohomekit/controller/ble/pairing.py
```diff
@@ -865,6 +865,38 @@ async def _close_while_locked(self) -> None:
...
+        the device will still be paired and the user
+        will have to un-pair it manually with a factory
+        reset.
+        """
+        if not self.accessories:
+            if self.description:
+                self._accessories_state = AccessoriesState(
+                    Accessories(), -1, self.broadcast_key
+                )
+                return self.description.name
```
bdraco, line 890: I think we could put this at the top and than pass to super() to reduce the duplicate code

bdraco:
Will also add some logic to make sure we don't give them back the short local name since currently there is no guard to prevent it from showing if we get it in an advertisement if we already have the complete local name

bdraco:
this will help as well https://github.com/esphome/aioesphomeapi/pull/298

bdraco:
retested pairing with local adapters as well to make sure that didn't regress 👍

PR #246, opened 2022-11-09
Bump aiocoap to 0.4.4

bdraco:
Our min required version was 0.4.1 and it doesn't have @roysjosh fix


Jc2k (approved):

bdraco:
Will push a tag shortly

PR #247, opened 2022-11-09
Reconnect if multiple CoAP PDUs go unanswered

roysjosh:


roysjosh:
Possibly fixes https://github.com/home-assistant/core/issues/80131

Jc2k:
Will this look like the device is disconnected when an operation is in flight?

roysjosh:
I've seen some accessories go unavailable and never recover. This is a guard against that. The PDU send/receive counters should be in sync except for the window of time between a request and response.

Jc2k:
Sure I get what this fixes, I'm just trying to imagine what could happen between the send and receive - could we solve one bug and introduce some races. E.g. if I toggle a button too quickly could the 2nd write to the device happen in parallel, see the device is disconnected, and kill the session, and then we need to re-auth.

Is there a potential for the user to see devices go unavailable randomly for a second, just because there is a request in flight?

I think we already have a timeout on `post_bytes` - maybe that should explicitly kill the session?

And maybe we should use an operation lock like BLE does to stop concurrent access?

roysjosh:
Ahh I see, good point. Hmm. Maybe the `post_bytes` exception handler would be a better location.

We could add that to be safe. Even if I found that Nanoleaf bulbs could handle >1 outstanding request, that may not be the same case for battery-powered devices.

roysjosh:
Approach this a different way. Network errors or timeouts increment a counter which counts towards being considered disconnected. This should solve situations where HAP+Thread accessories end up unavailable and never recover.

PR #248, opened 2022-11-10
Reduce some linear searches to find aid and iids

bdraco:
https://github.com/Jc2k/aiohomekit/pull/237/files#r1019213551


Jc2k (approved):

bdraco:
This one took a while to test even though its well covered as I wanted to be sure I didn't miss anything.

Tested with BLE and IP accessories to be sure 👍

PR #252, opened 2022-11-14
Add python 3.11 to the CI

bdraco:


bdraco:
blocked until HA updates via https://github.com/home-assistant/core/pull/78860

pyproject.toml
```diff
@@ -30,6 +30,7 @@ bleak-retry-connector = ">=2.3.0"
 orjson = ">=3.7.8"
 async-timeout = "^4.0.2"
 chacha20poly1305 = "^0.0.3"
+aiohttp = ">=3.8.3"
```
Jc2k, line 33: I thought this was just a dev dependency (to get asyncio support in pytest) - do we use it for something else now?

bdraco: It should just be dev.

bdraco: fixed. good catch

Jc2k (approved):

bdraco:
Will push a tag in a moment which should take care of the coap fix as well

PR #258, opened 2022-11-30
Use the same gatt write logic in _write_pdu as _read_signature

bdraco:
This speeds up writes a bit especially with the bluetooth proxies. 

No need to wait for a response if the device supports it because we are always waiting for the read response anyways.


Jc2k (approved):

bdraco:
Retested on mac os as well just to be sure

bdraco:
Will push a tag a bump hass in a moment

bdraco:
thanks!

PR #259, opened 2022-12-01
CoAP: Fix recovering from incorrect IPv6 address at start

Jc2k:
We only set the updated address when there was an encryption context. That meant we couldn't fix the IP if it was wrong at startup.

Fixes: https://community.home-assistant.io/t/homekit-accessory-protocol-hap-over-coap-udp-was-nanoleaf-essentials-bulb-via-thread-coap/335167/311?u=jc2k.


bdraco (approved):

Jc2k:
Cheers!

PR #261, opened 2022-12-01
Speed up BLE pair verify resume

bdraco:
We were reading the descriptor to get the iid every time 😢


Jc2k (approved):

bdraco:
thanks.  Will push a tag shortly

PR #263, opened 2022-12-02
Fix GSN fetching logic

bdraco:
The GSN will never change while we are connceted. We must increment it ourselves when we get a GATT notify, and we must only ever do it once per session since it does not actually change in the adv until we disconnect.

This fix ensures we do not reconnect again to the device when there is no change since we did not previously increment the GSN ourselves

We also refetch the broadcast encryption key when
the GSN rolls over as a roll over invalidates the
previous key


Jc2k (approved):

bdraco:
All good now 👍

bdraco:
Pushing a tag shortly

PR #268, opened 2022-12-03
Fix coap ip change handling

Jc2k:
Initial fix was tested in connection.py, and i messed it up moving it to here - there is no self.owner in paring!


Jc2k:
This is running on my prod now, seems ok.

bdraco (approved):

PR #270, opened 2022-12-06
Fix future warning about client.get_services

bdraco:
Fixes
```
/Users/bdraco/home-assistant/venv/lib/python3.10/site-packages/aiohomekit/controller/ble/pairing.py:813: FutureWarning: This method will be removed future version, use the services property instead.
  services = await self.client.get_services()
```


Jc2k (approved):

bdraco:
Thanks

PR #272, opened 2022-12-20
fix: bind to a specific local IPv6 address during pair verify

roysjosh:
This will prevent future traffic from picking the "best" address to communicate with an accessory. Current products tie sessions to the original IPv6 address and return a CoAP error code 4.04 when a new address attempts to continue the session.

It is likely that this could explode in fun, new ways: the DHCPv6 server returning a new address, the host selecting a new random SLAAC address, etc.


roysjosh:
Cross reference [aiocoap#298](https://github.com/chrysn/aiocoap/issues/298)

roysjosh:
I'm running this locally now. I've been seeing dozens of 4.04 errors a day. Let's see if this cuts that out.

roysjosh:
Wrap "get local IPv6 address" function in try/except. This prevents network errors from marking the device as not being provided by the integration.

This does seem to have eliminated frequent 404s from my environment.

jfroy:
This seems like a good fix, provided the library can react to local address changes. From the code, `_get_local_ipv6_addr`  lets the kernel pick a source address and interface to connect to the accessory and the library snapshots that and assumes it will be good "forever". I wonder if there's a performant way to monitor that route and reset if it becomes invalid (e.g. SLACC reconfiguration, etc). Though in general with ULA addresses that should not happen.

I wonder if the logic could be made to prioritize ULAs or link-local versus say an ISP publicly routable address that is more subject to change (especially when using privacy-preserving randomized addresses).

Jc2k:
The ip module has various attributes that could help with that: https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv6Address.is_private

Jc2k:
In terms of detecting failed connections, hoping we can get around to testing if the coap transport publishes a sane s# value in mdns. That number should go up if there is a state change that no one observed. So if we see it go up, we know we should restart the session.

Jc2k:
It's nice that we get a new source address every time we pair verify.

My instinct is that it'd be nice if we could have the events service bind to * but set the source address on outgoing traffic. But I don't think that helps much in practice - if the local ip changes the device won't know it and if the remote ip changes we handle it through mdns.

In terms of detecting our own ip changing, ideally I'd want to deter that to HA. Presumably they have to handle that to update their mdns record.

Jc2k (commented):
https://github.com/openthread/openthread/pull/1520 is another case where openthread code had to be mindful of source address.

Jc2k:
At the moment I'm tempted to merge the 4.04 handling and recovering only, as I'd like to try and understand the situations where the source address is wrong before deploying this work around. And I just don't think I'm going to get in front of a computer before the Jan release is out.

I'm struggling to think of scenarios where the source address flips if we have picked the thread ULA of the device as the destination address, the best one i came up with is if your HA loses its thread route but your home router doesn't. Default route would get it from HA to your main router, then your router delivers it to the thread router. But home routers don't generally accept thread routes out of the box, and would rely on your thread devices doing DHCP6, not just your thread router. That would be surprising as they are different L2 segments aren't they?

If we do go ahead with this work around I want to test what does happen if the ipv6 did address of HA did  change - hopeful any pending recvs would get cancelled, and bubble up asyncio into our code?

Jc2k:
2.4.2 handles `Code.NOT_FOUND` but doesn't contain the bind fix, going to put that into HA now. It also lowers the severity of a NOT_FOUND - something that can happen for valid reasons (like a session expiring while your border router is offline, battery changes, mains power cycling, etc) that we can automatically recover from shouldn't show up in the error log.

Jc2k:
And PR to land it in HA - https://github.com/home-assistant/core/pull/84700.

Hopefully I can find time with all my devices (test trashing the network etc ti test recovery) to test the full PR but can't say it's likely before mid jan.

roysjosh:
I upgraded HA which wiped out this full by-hand change. I don't "see" the reconnects any more from the perspective of the frontend but they are still happening w/o the full fix.

Jc2k:
Yep, still on my radar to loop back round to here and test out the full fix when I've got more time to try and break things.

PR #273, opened 2022-12-24
feat: update HAA characteristic types

jaredhobbs:
HAA changed the way they define custom characteristics. There is now only one custom characteristic and depending on the value written, there are four different actions that can be performed: Update, Setup, Reboot, and Reconnect WiFi.

Here's the corresponding HA PR: https://github.com/home-assistant/core/pull/84519


Jc2k (changes requested):
These are the same char, so we shouldn't have multiple names for them.

Jc2k:
(Thanks for taking a second pass at this by the way 👍)

jaredhobbs:
no longer needed, closing.

PR #274, opened 2022-12-28
Pad short uuid with 0's

Jc2k:
In CoAP we are seeing squished UUID's like `45E5011ECB4000A80FF2603DE`. These need padding to 32 chars with leading `0` before normalising.

E.g. `45E5011ECB4000A80FF2603DE` should become `00000004-5E50-11EC-B400-0A80FF2603DE`.

Via https://community.home-assistant.io/t/homekit-accessory-protocol-hap-over-coap-udp-was-nanoleaf-essentials-bulb-via-thread-coap/335167/346?u=jc2k

### DIFF
--- a/aiohomekit/uuid.py
+++ b/aiohomekit/uuid.py
@@ -55,6 +55,6 @@ def normalize_uuid(value: str) -> str:
     # Handle cases like 34AB8811AC7F4340BAC3FD6A85F9943B
     # Reject the rest
     try:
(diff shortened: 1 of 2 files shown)


Jc2k:
@bdraco fairly small fix for a thread device I want to get into next beta tag, but shared function with `ip` and `ble` so thought I better give you a heads up.

bdraco (approved):

Issue #276, opened 2023-01-09
Fetching friendly_name does a linear search


bdraco:
Its the most expensive part of the state update for some of my production devices.  Noticed while looking at some profiles for bluetooth
<img width="1423" alt="Screenshot 2023-01-08 at 2 23 00 PM" src="https://user-images.githubusercontent.com/663432/211226234-0eb17aff-bbfd-42b9-aac4-ee8e875720d1.png">

PR #278, opened 2023-01-12
Handling response as dict and list in case of failure in IpPairing#pu…

liebehentze:
I believe it solves the issue found over here:
https://github.com/home-assistant/core/issues/85400#issuecomment-1379381464

response for our shutter roller seems to be a single key dict as it can be seen in the log attached. The change fixes the errors I'm getting. There might be a different cause on this, but I can't keep track of all that.

[home-assistant_2023-01-12T11-12-01.387Z.log](https://github.com/Jc2k/aiohomekit/files/10401686/home-assistant_2023-01-12T11-12-01.387Z.log)

### DIFF
--- a/aiohomekit/controller/ip/pairing.py
+++ b/aiohomekit/controller/ip/pairing.py
@@ -299,9 +299,7 @@ class IpPairing(ZeroconfPairing):
             # If there is a response it means something failed so
             # we need to remove the listener update for the failed
             # characteristics.
-            if isinstance(response, dict) and "characteristics" in response:
-                response = response["characteristics"]
-            for characteristic in response:
+            for characteristic in response["characteristics"]:
                 aid, iid = characteristic["aid"], characteristic["iid"]
                 key = (aid, iid)
                 status = characteristic["status"]
--- a/aiohomekit/controller/ip/pairing.py
+++ b/aiohomekit/controller/ip/pairing.py
@@ -304,8 +304,8 @@ class IpPairing(ZeroconfPairing):
                 key = (aid, iid)
                 status = characteristic["status"]
                 status_code = to_status_code(status).description
-                if status_code != HapStatusCode.SUCCESS and key in listener_update:
-                    listener_update.pop(key)
+                if status_code != HapStatusCode.SUCCESS:
+                    listener_update.pop(key, None)
                 response_status[key] = {"status": status, "description": status_code}
 
         if listener_update:
--- a/aiohomekit/controller/ip/pairing.py
+++ b/aiohomekit/controller/ip/pairing.py
@@ -299,6 +299,8 @@ class IpPairing(ZeroconfPairing):
             # If there is a response it means something failed so
             # we need to remove the listener update for the failed
             # characteristics.
+            if isinstance(response, dict) and "characteristics" in response:
+                response = response["characteristics"]
             for characteristic in response:
                 aid, iid = characteristic["aid"], characteristic["iid"]
                 key = (aid, iid)
--- a/aiohomekit/controller/ip/pairing.py
+++ b/aiohomekit/controller/ip/pairing.py
@@ -304,7 +304,7 @@ class IpPairing(ZeroconfPairing):
                 key = (aid, iid)
                 status = characteristic["status"]
                 status_code = to_status_code(status).description
-                if status_code != HapStatusCode.SUCCESS:
+                if status_code != HapStatusCode.SUCCESS and key in listener_update:
                     listener_update.pop(key)
                 response_status[key] = {"status": status, "description": status_code}
 
--- a/tests/test_ip_pairing.py
+++ b/tests/test_ip_pairing.py
@@ -327,3 +327,12 @@ async def test_transport_property(pairing: IpPairing):
 
 async def test_polling_property(pairing: IpPairing):
     assert pairing.poll_interval == timedelta(seconds=60)
+
+
+async def test_put_characteristics_invalid_value(pairing: IpPairing):
+    aid, iid = (1, 2)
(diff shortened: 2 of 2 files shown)


Jc2k:
It looks like this bug is a regression introduced by https://github.com/Jc2k/aiohomekit/pull/237. We went from:

```python
        response_status: dict[tuple[int, int], dict[str, Any]] = {}
        if response:
            data = {
                (d["aid"], d["iid"]): {
                    "status": d["status"],
                    "description": to_status_code(d["status"]).description,
                }
                for d in response["characteristics"]
            }
            return data
```

to

```python
        response_status: dict[tuple[int, int], dict[str, Any]] = {}
        if response:
            # If there is a response it means something failed so
            # we need to remove the listener update for the failed
            # characteristics.
            for characteristic in response:
                aid, iid = characteristic["aid"], characteristic["iid"]
                key = (aid, iid)
                status = characteristic["status"]
                status_code = to_status_code(status).description
                if status_code != HapStatusCode.SUCCESS:
                    listener_update.pop(key)
                response_status[key] = {"status": status, "description": status_code}

        if listener_update:
            self._callback_listeners(listener_update)

        return 
```

Note that in the original code, we *always* looked at `response["characteristics"]`. So i think its wrong to make this fix conditional. I think we should *always* look at the characteristics key, just like we used to.

If i recall correctly, it's actually optional for the device to return a response on success. The spec says that if no error occurs, the device replies with a 204 and an empty body. So i think we are only hitting this when devices are off spec and return a status anyway, or they are genuinely failing. And that's why this error has mostly gone unnoticed for so long.

Jc2k:
Looking at your log, sometimes you get:

```
2023-01-12 12:11:47.890 DEBUG (MainThread) [aiohomekit.controller.ip.connection] 192.168.178.30: raw request: b'PUT /characteristics HTTP/1.1\r\nHost: 192.168.178.30\r\nContent-Length: 50\r\nContent-Type: application/hap+json\r\n\r\n{"characteristics":[{"aid":2,"iid":11,"value":0}]}'
2023-01-12 12:11:48.903 DEBUG (MainThread) [aiohomekit.controller.ip.connection] 192.168.178.30: raw response: bytearray(b'')
```

Although we don't log the status code, we did log that the body is empty.

We also see that sometimes you get:

```
2023-01-12 12:11:55.934 DEBUG (MainThread) [aiohomekit.controller.ip.connection] 192.168.178.30: raw request: b'PUT /characteristics HTTP/1.1\r\nHost: 192.168.178.30\r\nContent-Length: 52\r\nContent-Type: application/hap+json\r\n\r\n{"characteristics":[{"aid":2,"iid":11,"value":100}]}'
2023-01-12 12:11:56.450 DEBUG (MainThread) [aiohomekit.controller.ip.connection] 192.168.178.30: raw response: bytearray(b'{"characteristics":[{"aid":2,"iid":11,"status":-70402}]}')
```

Both of these match my theory, and I think the first request would have worked (beause of the `if response` guard), and the 2nd would have triggered the exception. This change will silence the exception for you, but your device is still intermittently reporting that its having a hard time obeying the command. That error code is for a comms error between a bridge and the device it controls iirc?

liebehentze:
Interesting, I think you have a way better overview on this. Do you want to update this fix or do you want me to do so?

Jc2k:
You already have a branch and a PR, and I don't have a dev environment with me, so if you could update the PR that would be great.

liebehentze:
Thanks for this look. I think you are right. 

The connectivity box (https://www.somfy.de/produkte/1870755/connectivity-kit) had also problems like this before using it without homekit directly with the vendor app.

I will update the PR and check if it resolves the problem at my local running ha instance.

liebehentze:
@Jc2k just easy as this?

Jc2k:
Yes, that makes it match what it was doing before #237 landed, though i'd still like @bdraco to take a look in case i've missed something.

bdraco:
Looks good to me.  Thanks for fixing.

Would it be possible to add a test for this?

liebehentze:
I was able to write a simple test without my device data to reproduce the bug. The test was failing before the fix with the exact same stacktrace on the issue referenced. Let me know if this fits your expectation.

aiohomekit/controller/ip/pairing.py
```diff
@@ -299,12 +299,12 @@ async def put_characteristics(
...
             # characteristics.
-            for characteristic in response:
+            for characteristic in response["characteristics"]:
                 aid, iid = characteristic["aid"], characteristic["iid"]
                 key = (aid, iid)
                 status = characteristic["status"]
                 status_code = to_status_code(status).description
-                if status_code != HapStatusCode.SUCCESS:
+                if status_code != HapStatusCode.SUCCESS and key in listener_update:
                     listener_update.pop(key)
```
bdraco, line 308:
```suggestion
                if status_code != HapStatusCode.SUCCESS:
                    listener_update.pop(key, None)
```

bdraco (approved):

bdraco:
Thanks @liebehentze

liebehentze:
@bdraco sure :). When is the next release planned including these changes?

Jc2k:
aiohomekit doesn't really have a release schedule, we just do ad-hoc releases as needed. I'd expect this to be released today or tomorrow and immediately submitted to HA.

liebehentze:
Sounds good

bdraco:
I was planning on looking into https://github.com/Jc2k/aiohomekit/issues/276 

If I can come up with a solution for that, I'll do a release after it, or otherwise do one if I can't

liebehentze:
Good job, thanks a lot

PR #279, opened 2023-01-13
Avoid linear searches to get services by service_type

### LINKED ISSUE #276: Fetching friendly_name does a linear search

bdraco:
Since most lookups are by service_type we now store them in a dict under the hood to avoid linear searches

fixes #276


Jc2k (approved):

bdraco:
Thanks.  Will push a tag shortly

PR #285, opened 2023-02-17
Add enums for air purifier states

dcmeglio:
Add enums for current and target air purifier states.


dcmeglio:
I feel like something needs to be done in data.py to map the characteristic to an enum but the file says `# AUTOGENERATED, DO NOT EDIT` so I am assuming you have some tool that will handle this for me? If not let me know what the process is to get this file regenerated.

Jc2k:
Yes I'll take care of that.

PR #286, opened 2023-04-03
Bump zeroconf to 0.54.0 to fix handling server name changes

bdraco:
Removes the need for _first_non_link_local_address since zeroconf now guarantees LIFO order for addresses


Jc2k (approved):

bdraco:
Thanks. Will push a tag and bump hass shortly

I think the other change is safe?

PR #292, opened 2023-06-01
Only lookup the accessory_information service once

### LINKED ISSUE #289: `first` to get accessory_info at startup is unexpectedly expensive
Screenshot 2023-05-30 at 9 21 49 AM

bdraco:
There can only be one of these so we can cache the lookup once we do it once

fixes #289


Jc2k (approved):

bdraco:
Ran overnight ok 👍

bdraco:
thanks

PR #296, opened 2023-06-24
Add enum values for air quality

agoode:
See https://github.com/home-assistant/core/issues/94754.


Jc2k:
aiohomekit/model/characteristics/data.py needs updating as well before this can be used in HA like the Thread enums. I'll do that before release, there are some other changes pending first.

PR #298, opened 2023-07-10
Handle BLEDevice name being `None` during discovery

bdraco:
will fix (but does not close) https://github.com/home-assistant/core/issues/96258

### DIFF
--- a/tests/test_controller_ble.py
+++ b/tests/test_controller_ble.py
@@ -1,4 +1,6 @@
+from __future__ import annotations
 from typing import Any
+
 from unittest.mock import MagicMock
 
 from bleak.backends.device import BLEDevice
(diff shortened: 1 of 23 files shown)


Jc2k (approved):

tests/test_controller_ble.py
```diff
@@ -0,0 +1,110 @@
+from __future__ import annotations
```
bdraco, line 1: Might as well start somewhere with the tests here

bdraco:
Doing an unpair - repair to be sure everything is ok and than will merge + release

bdraco:
All works but need to run to airport so will release when I get back from dropping off

bdraco:
All good after retesting

PR #299, opened 2023-07-11
Fix ble controller discovery where old name is None

bdraco:
The fix in #298 was not sufficient


bdraco:
Sorry for the churn as I missed this in #298

PR #197, opened 2022-10-19
Add support for encrypted notifications

bdraco:
We now try to get the char change via an encrypted notification instead of a disconnected
event when possible.  This avoids the need to connect to the accessory for changes that
are made to it from outside of aiohomekit which drastically speeds up updates and frees
up BLE connection slots.

PR #301, opened 2023-07-16
Remove workaround to populate missing handles

bdraco:
#197 added a check for missing handles if the user was running an older version. As its been a few months, everyone should have the missing handles populated now, and since this check is rather expensive to run every operation, we should remove it.


Jc2k (approved):

bdraco:
thanks

bdraco:
I'll push a release soon. Didn't have time to do it earlier since I'm traveling

PR #307, opened 2023-07-22
Reduce memory churn and copy while decoding large payload

bdraco:
I was testing with some cameras and noticed the image encryption overhead was showing up in the profile

- Implement the todo to not return `False`


bdraco:
I thought it was caused by this PR, but pressing identify on the eve energy causes a disconnect

Happens on main as well

bdraco:
problem was with the esp32 but this code

bdraco:
coap uses cryptography directly so no changes there

Jc2k (approved):

Jc2k:
😍 for getting rid of that return false.

bdraco:
thanks

PR #309, opened 2023-07-22
Fix cancellation race in IP controller

bdraco:
related issue https://github.com/home-assistant/core/issues/93213


aiohomekit/controller/ip/connection.py
```diff
@@ -545,9 +551,6 @@ async def _reconnect(self):
                     # Authentication errors should bubble up because auto-reconnect is unlikely to help
                     raise
 
-                except asyncio.CancelledError:
-                    return
-
```
bdraco, line 550: Newer python changed CancelledError to inherit from base exception so we should not catch it here and let it raise instead otherwise this leads to odd behavior since it blocks cancellation but doesn't decrement the cancellation counter with uncancel

aiohomekit/controller/ip/connection.py
```diff
@@ -563,16 +566,14 @@ async def _reconnect(self):
...
                 interval = min(60, 1.5 * interval)
-                self._reconnect_wait_task = asyncio.ensure_future(
-                    asyncio.sleep(interval)
-                )
-
+                self._reconnect_future = self._loop.create_future()
                 try:
-                    await self._reconnect_wait_task
-                except asyncio.CancelledError:
+                    async with interrupt(self._reconnect_future, ConnectionReady, None):
```
bdraco, line 571: This library avoids creating a task which avoids a race

Jc2k (approved):
Nice!

aiohomekit/controller/ip/connection.py
bdraco: I need to check this since I think part of the reason it was added was to prevent cancelation from bubbling up on anything that awaits the task.  We likely need to shield higher up for that though

bdraco:
I had to tweak it a bit since we await it via ensure_conncted as well and we don't want a timeout to cancel the inner.

This also fixes shutdown having to wait for the timeout because cancellation was ignored in the task

bdraco:
there are actually two time outs for the exact same time so there is another race here

bdraco:
nevermind, the other timeout is fine since its wrapping the `ensure_connected` which goes into the background if it cannot connect. Its just confusing that they are the same

Jc2k:
That background reconnect thing seemed like a good idea at the time (before async zeroconf etc). But if you feel it's adding complexity, maybe we could get rid. Polling will naturally retry, zeroconf updates should trigger a retry also. And we could add zeroconf s# support (if it's present and not hardcodes to 1 then it seems to work like BLE) to trigger reconnects.

bdraco:
There is also another race here since the add callback done is delivered via a call_soon so we only see it on the next iteration which means it can set the task to none when it's already started a new one

bdraco:
I think the background reconnect is still needed for users with unreliable mdns

aiohomekit/controller/ip/connection.py
```diff
@@ -292,9 +301,18 @@ async def _stop_connector(self):
...
+        # since the connector will be canceled when the connection is closed.
+        #
+        # Cancellation of the connector will still happen but we won't
+        # propagate it higher in the stack.
+        try:
+            await self._connector
+        except asyncio.CancelledError:
+            pass
+        # self._connector being set to None is the responsibility of the
+        # done_callback in _start_connector
```
bdraco, line 315:
```suggestion
```

aiohomekit/controller/ip/connection.py
```diff
@@ -280,8 +282,15 @@ async def ensure_connection(self):
         Otherwise, start a reconnection and wait for it.
         """
         if self._start_reconnecting():
+            # If we are running under a timeout, we still need to shield the
+            # connector task so it continues to run if the timeout is hit.
             await asyncio.shield(self._connector)
 
+        if self._last_connector_error:
+            raise self._last_connector_error
```
bdraco, line 290: We now raise the last connection error if there was a problem connecting since all we ever told the user was it timed out otherwise

aiohomekit/controller/ip/pairing.py
```diff
@@ -139,6 +139,10 @@ async def _ensure_connected(self):
             raise AccessoryDisconnectedError(
```
bdraco, line 145: We currently unconditionally wait for the timeout on failure.

That would be a lot more to refactor so I worked on getting the error message to be better instead.

bdraco: To test this I changed the ip address to start with 224 in the config entry
Screenshot 2023-07-22 at 2 55 47 PM


I think this will help a lot with issue reports

bdraco: Previously it would always say it timed out

bdraco:
I did a bit of chaos testing with this by having a smart plug turn off a homekit smart plug every 90 seconds and while the homekit plug is flipping on/off every second

Everything works now with no reconnect failures

aiohomekit/controller/ip/pairing.py
bdraco: Also pairing failures now show up here as well

bdraco: Although the exception messages are pretty bad

PR #310, opened 2023-07-22
Fix some missed False checks from converting decrypt to raise an exception

bdraco:


Jc2k (approved):

bdraco:
grepped a bit more. pretty sure I got them all now

PR #318, opened 2023-08-06
Make a constant for EVENT_CHARACTERISTICS

bdraco:
related issue https://github.com/home-assistant/core/issues/97871
intended use case https://github.com/home-assistant/core/pull/97877


bdraco:
I hate to add a constant, but this is a legitimate special case

Jc2k (approved):

bdraco:
Tried to abstract this a bit but the complexity ended up not being worth the trade off.

Constant seems to be the simplest solution.

Will push a release shortly

PR #322, opened 2023-08-25
Make HomeKitService slotted

bdraco:
We tend to have a lot of these in memory if there are a lot of HomeKit devices on the network (even if they are not paired to HKC)

### DIFF
--- a/aiohomekit/zeroconf.py
+++ b/aiohomekit/zeroconf.py
@@ -21,9 +21,7 @@ from abc import abstractmethod
 import asyncio
 from collections.abc import AsyncIterable
 from dataclasses import dataclass
-from functools import partial
 import logging
-import sys
 
 import async_timeout
 from zeroconf import (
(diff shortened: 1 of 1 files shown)


aiohomekit/zeroconf.py
```diff
@@ -57,8 +59,13 @@
 
 logger = logging.getLogger(__name__)
 
+if sys.version_info[:2] < (3, 10):
```
Jc2k, line 62: I am happy to follow HA versions here and drop < 3.10 to avoid the if, wdyt?

bdraco: That works for me. We could drop the asyncio.timeout/async_timeout fallback in utils.py as well

bdraco: https://github.com/Jc2k/aiohomekit/pull/325

bdraco: 3.11 only like HA might too aggressive. not sure

bdraco: I think it probably is for anyone else using this.... is anything else using this?

Jc2k: No one's told us if they are, and we know that the CLI pairing broke during the multi transport changes so they can't really be using that.

bdraco:
That makes things very simple here.

Thanks.

PR #324, opened 2023-08-28
Hold strong references to tasks

bdraco:
> https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

> Important Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection


Jc2k (approved):

bdraco:
This one will likely fix some unexplained failures

PR #327, opened 2023-08-28
Reverting adding typing return to clamp_enum_to_char

bdraco:
This is going to require some more refactoring to make it work

revert it for now

partial revert of #323


bdraco:
I guess we could at least make it `set[Any]` for now

PR #329, opened 2023-09-02
Ensure we resolve services again after the debounce timeout

bdraco:
Since we held on to the AsyncServiceInfo for as long as the service was alive, we never respected the TTL of the DNSAddress records which meant we effectively held on to them forever as long as the device was on the network

related issue https://github.com/home-assistant/core/issues/86029


Jc2k (approved):

bdraco:
trace looks good

_async_resolve_later
_handle_service

bdraco:
TODO 

- [ ] change ip of switchbot plug and make sure its picked up

bdraco:
verified ip change works on switchbot plug mini

bdraco:
no errors in the log on restart, works again on restart

PR #330, opened 2023-09-09
Drop async_timeout on py<3.11

bdraco:
We can drop the async-timeout dep on py3.11+

There was one still left in zeroconf, but it is now switched out


Jc2k (approved):

bdraco:
thanks

PR #331, opened 2023-09-09
Create futures with loop.create_future()

bdraco:
https://docs.python.org/3/library/asyncio-future.html#asyncio.Future

> The rule of thumb is to never expose Future objects in user-facing APIs, and the recommended way to create a Future object is to call loop.create_future(). This way alternative event loop implementations can inject their own optimized implementations of a Future object.


Jc2k (approved):

bdraco:
Thanks. Will do a release shortly. Nothing to interesting here but would rather have a release in case there is a regression (don't except one since I've made similar changes in many other libs...but errors happen) so its easier to test later

PR #335, opened 2023-09-25
More workarounds for Schlage Encode Plus

Jc2k:
* Another place where "linked_services" contains a 0
* The linked service TLV has length set to 255, but at end of byte stream. Return the data what we do have and ignore the length.


Jc2k:
See https://github.com/home-assistant/core/issues/100160

bdraco (approved):

Jc2k:
Cheers. Will bump aiohomekit in Ha later.

PR #336, opened 2023-09-27
Try harder to avoid missing button presses

### LINKED ISSUE #319: BLE Events might still be read when processing disconnected notifications
I was testing with the onvis multi button

If it press one button and than another quickly, it will generate an encrypted notify and than a disconnected event

We miss the 2nd event because we don't poll the event char

bdraco:
maybe fixes #319


bdraco:
This fixes the missed button presses but now I get duplicates

So if I do

4, 5, 3

4 is broadcast as adv and fired
5 isn't and generates a disconnected event

It connects and reads 4,5,3 so 4 gets fired again

PR #338, opened 2023-10-19
Small cleanups to IP connection

bdraco:
- Add more typing
- Add some doc strings
- Avoid async_timeout overhead when waiting for a future

### DIFF
--- a/aiohomekit/controller/ip/connection.py
+++ b/aiohomekit/controller/ip/connection.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from async_interrupt import interrupt
 
@@ -54,10 +54,12 @@ class ConnectionReady(Exception):
 
 
 class InsecureHomeKitProtocol(asyncio.Protocol):
-    def __init__(self, connection):
+    """An asyncio.Protocol implementation for HomeKit connections."""
+
+    def __init__(self, connection: HomeKitConnection) -> None:
         self.connection = connection
         self.host = ":".join((connection.host, str(connection.port)))
-        self.result_cbs = []
+        self.result_cbs: list[asyncio.Future[HttpResponse]] = []
         self.current_response = HttpResponse()
         self.loop = asyncio.get_running_loop()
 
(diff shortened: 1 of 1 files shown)


Jc2k (commented):
Body is bytes|None

aiohomekit/controller/ip/connection.py
```diff
@@ -323,11 +340,13 @@ async def get(self, target):
...
         )
 
-    async def get_json(self, target):
+    async def get_json(self, target: str) -> dict[str, Any]:
         response = await self.get(target)
         return hkjson.loads(response.body)
 
-    async def put(self, target, body, content_type=HttpContentTypes.JSON):
+    async def put(
+        self, target: str, body: Any, content_type=HttpContentTypes.JSON
```
Jc2k, line 348: bytes (none optional here)

aiohomekit/controller/ip/connection.py
```diff
@@ -370,7 +389,9 @@ async def put_json(self, target, body):
 
         return parsed
 
-    async def post(self, target, body, content_type=HttpContentTypes.TLV):
+    async def post(
+        self, target: str, body: Any, content_type=HttpContentTypes.TLV
```
Jc2k, line 393: And here

Jc2k (changes requested):
Looks good! Thanks. I think we can make body types tighter but otherwise all goos

aiohomekit/controller/ip/connection.py
bdraco: 38cdf65d6aaa80edc21a02d9ac432a9b1daae5a7

aiohomekit/controller/ip/connection.py
bdraco: 38cdf65d6aaa80edc21a02d9ac432a9b1daae5a7

bdraco: I guess this one needs to not be none

bdraco: 073049de3d622ef50d1e91bcecfb5cc6f01735fc

Jc2k (approved):
Looks good! Thanks. I think we can make body types tighter but otherwise all good

bdraco:
Ran overnight all good 👍

PR #339, opened 2023-10-20
Make process_changes return a set of what actually changed

bdraco:
I realized we still write state too frequently. When we do the poll if the value is the same we still callback the entity.

Will be used in https://github.com/home-assistant/core/pull/102370

### DIFF
--- a/aiohomekit/model/__init__.py
+++ b/aiohomekit/model/__init__.py
@@ -27,6 +27,7 @@ from aiohomekit.uuid import normalize_uuid
 from . import entity_map
 from .categories import Categories
 from .characteristics import (
+    EVENT_CHARACTERISTICS,
     Characteristic,
     CharacteristicFormats,
     CharacteristicPermissions,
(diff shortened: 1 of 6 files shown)


aiohomekit/model/__init__.py
```diff
@@ -407,16 +407,29 @@ def has_aid(self, aid: int) -> bool:
...
+    ) -> set[tuple[int, int]]:
+        """Process changes from a HomeKit controller.
+
+        Returns a set of the changes that were applied.
+        """
+        changed: set[tuple[int, int]] = set()
+        for aid_iid, value in changes.items():
+            (aid, iid) = aid_iid
             if not (char := self.aid(aid).characteristics.iid(iid)):
                 continue
```
Jc2k, line 422: Should we just always add it to changed if it's one of the event characteristics?

bdraco: Yes we will need to do that

bdraco: https://github.com/Jc2k/aiohomekit/pull/339/files#diff-5443a6f050cba5fd14b7604e81e8139f66401a7f2eb74c8e8f396837127b2b3cR425

Jc2k (approved):

bdraco:
thanks, will release shortly

PR #340, opened 2023-10-21
Avoid creating TimerHandle when already connected

bdraco:
The check to see if a device was already connected was inside an asyncio.timeout block. This has the overhead of creating and canceling a timer in the event loop every time it was run. Add a guard before the timeout to avoid this when already connected


Jc2k (approved):

bdraco:
thanks

bdraco:
Will push a release to pickup your `md=` fix as well

PR #342, opened 2023-10-31
Ensure all transports can handle an iterable in get_characteristics

bdraco:
- Add some missing typing

fixes home-assistant/core#103103

- coap: needed an subscriptable list
- ble: didn't care about order
- ip: didn't care about order and always converted to a set


Jc2k (approved):

bdraco:
Will push a release and bump HA shortly

PR #349, opened 2023-12-10
Add support for happy eyeballs (RFC 8305)

### LINKED ISSUE #348: Implement happy eye balls with aiohappyeyeballs
Much can be borrowed from https://github.com/esphome/aioesphomeapi/pull/789

But the main issue here is aiohomekit doesn't have a concept of multiple addresses

bdraco:
Adds Happy Eyeballs support (https://datatracker.ietf.org/doc/html/rfc8305)

--

This PR is a first step as I need to make sure this library works for the other projects as well. Its likely to sit for a bit while I test the library with all the projects that need it so there aren't any need to do breaking changes before all the other projects start consuming it:
- https://github.com/aio-libs/aiohttp/pull/7954
- https://github.com/esphome/aioesphomeapi/pull/789

--

This should fix all the cases where

- We have dual stack ipv4/ipv6
- There are multiple IPs to connect to and we don't know which one is the correct one

closes #348

likely fixes https://github.com/home-assistant/core/issues/102901


bdraco:
I'm seeing duplicate IPv6 addresses

I think its a bug in zeroconf.

Going to have to fix that first

bdraco:
bug in zeroconf fixed in 0.128.2+

Will bump the min version here and the dep in HA

bdraco:
tests fail with new zeroconf because `_set_properties` is no longer exposed.

Will re-expose it for now https://github.com/python-zeroconf/python-zeroconf/pull/1327

.github/workflows/ci.yml
```diff
@@ -130,6 +130,6 @@ jobs:
 
     - name: Run pytest
       shell: bash
-      run: poetry run python -m pytest --cov=. --cov-report=xml
+      run: poetry run python -m pytest -vvvs --cov=. --cov-report=xml --timeout=60
```
bdraco, line 133: need to take this out once I figure out why the windows tests stall

aiohomekit/controller/ip/connection.py
```diff
@@ -477,7 +495,7 @@ async def request(
 
         # WARNING: It is vital that a Host: header is present or some devices
         # will reject the request.
-        buffer.append(f"Host: {self.host}")
+        buffer.append(f"Host: {self.connected_host or self.hosts}:{self.port}")
```
Jc2k, line 498: I was wondering if this would leave to invalid host headers (which can crash some of the crappier vendors), but we shouldn't be able to get here if there is no connected host? We'd be generate a http frame that we didn't sent? So is this needed? Or can we bail out earlier?

aiohomekit/controller/ip/connection.py
```diff
@@ -663,14 +697,14 @@ async def _connect_once(self):
...
             pairing = self.owner
             try:
-                if self.host != pairing.description.address:
+                if set(self.hosts) != set(pairing.description.addresses):
                     logger.debug(
                         "%s: Host changed from %s to %s",
                         pairing.name,
-                        self.host,
+                        self.hosts,
                         pairing.description.address,
```
Jc2k, line 705: .addresses?

Jc2k (commented):
This is looking good. Do you think it would be safe to allow ipv6 link local whilst working on this? It's much more likely to be a valid addresss than an ipv4 link local.

aiohomekit/controller/ip/connection.py
bdraco: It shouldn't be possible to get here without a connected host actually

We can generate the host header when it connects

for ipv6 we need brackets as well anyways

bdraco: 3193fdad7fc2eb99216b9d465ebc2cadee591f9b

aiohomekit/controller/ip/connection.py
bdraco: 166aeb549e245079456bf2bdb847168b97d64f4c

bdraco:
> This is looking good. Do you think it would be safe to allow ipv6 link local whilst working on this? It's much more likely to be a valid addresss than an ipv4 link local.

I fixed up zeroconf to pass the scope_id correctly so we could make ipv6 link local work with some more work but it would need a lot more refactoring to be able to split the `%<scope_id>` off of the addresses so I kept them excluded for now.   Something for the future.

Jc2k (approved):

bdraco:
Thanks.

I've just finished the code for all 3 of these PRs.

I'm doing testing aiohomekit and aioesphomeapi but I don't want to merge until I'm sure there won't be any breaking changes in the lib

bdraco:
Thanks. 

Happy with the other libs, will push a release shortly

PR #353, opened 2024-01-04
Remove port from Host header

bdraco:
related issue https://github.com/home-assistant/core/issues/107137

When we added happy eyeballs, I added the port, but this seems to cause problems for some devices


Jc2k (approved):
Cheers

PR #354, opened 2024-01-13
Handle cancellation while waiting for an encrypted IP response

bdraco:
Only fixes the IP controller.  BLE and coap likely have similar issues.  I will fix BLE as well but I don't use coap so no way to test that one

https://github.com/home-assistant/core/issues/97888


aiohomekit/controller/ip/connection.py
```diff
@@ -114,12 +112,19 @@ async def send_bytes(self, payload: bytes) -> HttpResponse:
...
         try:
+            self.transport.write(payload)
             return await result
-        except asyncio.TimeoutError:
-            timeout_expired = True
+        except (asyncio.TimeoutError, BaseException) as ex:
+            # If we get a timeout or any other exception then we need to
+            # close the connection as we are now out of sync with the device
+            # and any future requests will fail since the encryption counters
+            # will be out of sync.
```
bdraco, line 121: We might be able to redo the pair verify and start over but I doubt devices handle that well so I think the safest thing to do is close and reopen to ensure good state

Jc2k (approved):

bdraco:
thanks. will push a release shortly if everything passes on 3.12 as well

PR #358, opened 2024-02-02
Recover CoAP pairing when remote device stops responding

Jc2k:
Prior to this change if a remote device stopped responding (e.g. packet loss) we would never recover - we'd just keep incrementing the encryption counter making it further and further away from the truth.

The only way to recover from this right now is to restart HA which obviously resets the encryption session.

This patch can't fix the buggy device, unstable network or packet loss that lead to a timeout in the first place but it can help things recover without needing a full HA restart.

Tested by a community member with a very unstable set of Nanoleaf devices (multiple devices go offline at same time, and with this patch they actually recover by themselves).

### DIFF
--- a/aiohomekit/controller/coap/connection.py
+++ b/aiohomekit/controller/coap/connection.py
@@ -170,6 +170,10 @@ class EncryptionContext:
                 async with asyncio_timeout(timeout):
                     response = await self.coap_ctx.request(request).response
             except (NetworkError, asyncio.TimeoutError):
+                logger.debug("%s: Did not receive a reply; end of session.", self.uri)
+                if self.coap_ctx:
+                    await self.coap_ctx.shutdown()
+                    self.coap_ctx = None
                 raise AccessoryDisconnectedError("Request timeout")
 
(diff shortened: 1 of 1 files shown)


bdraco (approved):

Jc2k:
Cheers

codyc1515:
Do you have an issue # on here where someone faced this issue that you expect it would be resolved? i.e. is this just going to fix (workaround) the encryption counter issue or we would expect it to also fix ones like below

```
Logger: aiohomekit.controller.coap.pdu
Source: components/homekit_controller/connection.py:283
First occurred: 13:56:35 (6 occurrences)
Last logged: 15:33:15

Transaction 0 failed with error 6 (Invalid request
```

Jc2k:
No, it's just something we worked through on discord.

It will just stop (some?) devices getting stuck offline.

I seem to remember your error was known to the original dev of coap support, and it was harmless.

codyc1515:
Mint. I see no Unavailable messages in my logs anymore, so was curious what changed. Thank you so much.

PR #363, opened 2024-02-22
Small speed up to fetch characteristic values

bdraco:
We can avoid most of the body of this function for the common case.  We call this a lot when adding entities

Screenshot 2024-02-22 at 2 15 09 AM

### DIFF
--- a/aiohomekit/model/characteristics/characteristic.py
+++ b/aiohomekit/model/characteristics/characteristic.py
@@ -192,7 +192,14 @@ class Characteristic:
 
     @property
     def value(self) -> Any:
-        extra_data = characteristics.get(self.type, {})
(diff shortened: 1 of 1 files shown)


Jc2k:
Tangentially related, but do you think there is a nicer way to replace this entirely so we can do more strict type checking in HA? Was thinking about read_int(), read_string(), etc. I guess they could return self._value pretty directly and avoid the body too. But not sure about returning enums of different types and different TLV structs with that approach. Can sort of imagine a read_struct that took a class, and the return value was generic over that parameter.

If we did that is there a performant way (or, do we disable asserts at runtime in HA?) to check self.format matches the read function getting called?

Jc2k (approved):

PR #364, opened 2024-02-22
Avoid creating tasks to process zeroconf updates

bdraco:
Since ip pairing and coap pairings do nothing in _process_disconnected_events, make it a normal function instead and have the ble pairing create tasks as needed


Jc2k:
LGTM

Jc2k (approved):

PR #370, opened 2024-04-26
Implement async_reachable to see if a device is reachable on the network

bdraco:
For zeroconf based devices we check if the A/AAAA records are still published. This means we will know if the device is offline within 120s since A/AAAA records typically have a TTL of 120s

For BLE based devices we will check if the BLE device is in range and seen by the adapter. This will vary by BLE adapter, but it will likely mean we will know within a few minutes as well


bdraco:
I need to write a test for the zeroconf case

bdraco:
Keep getting distracted by beta issues so will have to come back to writing the tests later

bdraco:
I've been running this for a few days, and all seems well here

bdraco:
No crashes on the device now. I think this solve the problem

bdraco:
I haven't had one crash on the device since implementing this. This puts us much more in-line with the native HomeKit behavior.

bdraco:
Retested with Jc2k SO_KEEPALIVE idea. All good 👍

PR #371, opened 2024-06-17
Eliminate use of globals to generate ids

bdraco:
BREAKING CHANGE:

`Accessory` now requires an `aid`
`Accessory.create_with_info` now requires an `aid`


Jc2k (approved):

bdraco:
need to fix the CI. Will do that in another PR first after lunch

bdraco:
also I think https://github.com/Jc2k/aiohomekit/pull/370 is ready to go.  I don't have any more crashes on the devices mentioned in the linked issue anymore

PR #376, opened 2024-08-11
Fix IP controller not reconnecting right away

bdraco:
There were two problems that prevented the IP transport from reconnecting

1. When the connection was lost, _start_connector would not try to reconnect because self.transport and self.protocol were still set which meant the connection was not full torn down and it thought it was still connected. This is fixed by clearing transport and protocol before calling _start_connector.  This meant we would never reconnect until a poll attempt which restarted the connector.

2. If a request was inflight the connection would get re-established and than the concurrency lock would prevent the pair-verify from happening. This is fixed by resolving all the futures that are waiting when the connection is lost.

### DIFF
--- a/aiohomekit/controller/ip/connection.py
+++ b/aiohomekit/controller/ip/connection.py
@@ -86,6 +86,7 @@ class InsecureHomeKitProtocol(asyncio.Protocol):
 
     def connection_lost(self, exception):
         self.connection._connection_lost(exception)
+        self._cancel_pending_requests()
 
     def _handle_timeout(self, fut: asyncio.Future[Any]) -> None:
         """Handle a timeout."""
@@ -151,6 +152,9 @@ class InsecureHomeKitProtocol(asyncio.Protocol):
         return False
 
     def close(self):
+        self._cancel_pending_requests()
+
+    def _cancel_pending_requests(self) -> None:
         # If the connection is closed then any pending callbacks will never
         # fire, so set them to an error state.
         while self.result_cbs:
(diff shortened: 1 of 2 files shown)


aiohomekit/controller/ip/connection.py
```diff
@@ -736,6 +743,11 @@ async def _connect_once(self):
                 pass
 
         await super()._connect_once()
+        logger.debug(
+            "SecureHomeKitConnection established to %s:%s",
+            self.connected_host,
+            self.port,
+        )
```
bdraco, line 750:
```suggestion
```

aiohomekit/controller/ip/connection.py
```diff
@@ -616,7 +621,9 @@ async def _connect_once(self) -> None:
             self.host_header = f"Host: [{connected_host}]"
         else:
             self.host_header = f"Host: {connected_host}"
+
```
bdraco, line 624:
```suggestion
```

aiohomekit/controller/ip/connection.py
```diff
@@ -616,7 +621,9 @@ async def _connect_once(self) -> None:
...
         if self.owner:
+            logger.debug("Connection made to %s:%s", self.hosts, self.port)
```
bdraco, line 626:
```suggestion
```

aiohomekit/controller/ip/connection.py
```diff
@@ -597,6 +601,7 @@ async def _connect_once(self) -> None:
                 raise TimeoutError("Timeout") from last_exception
             raise ConnectionError(str(last_exception)) from last_exception
 
+        logger.debug("Connected established to %s:%s", self.hosts, self.port)
```
bdraco, line 604:
```suggestion
```

Jc2k (approved):

bdraco:
Thanks.  I'll get a release done shortly

PR #377, opened 2024-08-17
Fix race when during timeout of IP connection request

bdraco:
If the requests hits th 30s timeout, but responds or times out at just the right time before the during the connection teardown, the future will already be resolved.


Jc2k (approved):

bdraco:
I’m traveling this weekend but I’ll do a release when I get back to the room

PR #378, opened 2024-08-25
Move supports_subscribe to AbstractPairing

bdraco:
We have a check to see if subscribe is broken but checking it depends on the pairing type. Make sure all pairings support this.

related issue https://github.com/home-assistant/core/issues/124099


bdraco:
In https://github.com/home-assistant/core/issues/124099 we see the connection getting dropped. However the full log is missing so its hard to tell what is really going on there.

PR #382, opened 2024-10-18
Add current and target fan state enums

Xeio:
Adding per https://github.com/home-assistant/core/pull/128618

### DIFF
--- a/aiohomekit/model/characteristics/const.py
+++ b/aiohomekit/model/characteristics/const.py
@@ -141,7 +141,8 @@ class SwingModeValues(enum.IntEnum):
 
 class CurrentFanState(enum.IntEnum):
     """Possible values current fan state.
-    https://developer.apple.com/documentation/homekit/hmcharacteristicvaluecurrentfanstate"""
+    https://developer.apple.com/documentation/homekit/hmcharacteristicvaluecurrentfanstate
+    """
 
     INACTIVE = 0
     IDLE = 1
--- a/aiohomekit/model/characteristics/const.py
+++ b/aiohomekit/model/characteristics/const.py
@@ -139,6 +139,23 @@ class SwingModeValues(enum.IntEnum):
     ENABLED = 1
 
 
+class CurrentFanState(enum.IntEnum):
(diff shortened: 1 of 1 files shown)


aiohomekit/model/characteristics/const.py
```diff
@@ -139,6 +139,23 @@ class SwingModeValues(enum.IntEnum):
     ENABLED = 1
 
 
+class CurrentFanState(enum.IntEnum):
+    """Possible values current fan state.
+    https://developer.apple.com/documentation/homekit/hmcharacteristicvaluecurrentfanstate"""
```
Jc2k, line 144:
```suggestion
    https://developer.apple.com/documentation/homekit/hmcharacteristicvaluecurrentfanstate
    """
```

aiohomekit/model/characteristics/const.py
```diff
@@ -139,6 +139,23 @@ class SwingModeValues(enum.IntEnum):
...
+
+    INACTIVE = 0
+    IDLE = 1
+    ACTIVE = 2
+
+
+class TargetFanState(enum.IntEnum):
+    """Possible values target fan state.
+    https://developer.apple.com/documentation/homekit/hmcharacteristicvaluetargetfanstate"""
```
Jc2k, line 153:
```suggestion
    https://developer.apple.com/documentation/homekit/hmcharacteristicvaluetargetfanstate
    """
```

Jc2k:
Cheers. Just had to fix a black nag. CI should push a new aiohomekit to pypi in a few minutes.

Xeio:
Should I have also added this to \_\_init__.py? Just realized when I'm going to import it I have to use a different namespace

image

Jc2k:
Oh good shout, let's do that.

Xeio:
While I'm doing that... out of curiosity should I also have the word "Values" on the end? I sort of realized almost all the enums in that const.py have it on the name.

Jc2k:
That sounds like something past me thought was a good idea, let's do that.

PR #383, opened 2024-10-18
Add fan state enums to __init__

Xeio:
Per some additional details from #382


Jc2k:
They need adding to __all__ to shut the linter up.

PR #386, opened 2024-11-12
Fix pairings being loaded before transports

soumya92:
This fixes an issue I observed locally, where running `aiohomekitctl` with any existing IP pairing would throw an error:

```
controller.py:0185 ERROR Skipped pairing: Transport IP not supported. See setup.py for required dependencies.
```

After tracing, I discovered this is because the pairings were loaded before the transports, and moving the load_data call to this new location seemed to have fixed it for me.

### DIFF
--- a/aiohomekit/__main__.py
+++ b/aiohomekit/__main__.py
@@ -52,12 +52,6 @@ async def get_controller(args: argparse.Namespace) -> AsyncIterator[Controller]:
         char_cache=CharacteristicCacheFile(charmap_path),
     )
 
-    try:
-        controller.load_data(args.file)
-    except Exception:
-        logger.exception(f"Error while loading {args.file}")
-        raise SystemExit
-
     async with zeroconf:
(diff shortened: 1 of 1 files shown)


aiohomekit/__main__.py
```diff
@@ -70,6 +64,11 @@ async def get_controller(args: argparse.Namespace) -> AsyncIterator[Controller]:
         )
 
         async with controller:
+            try:
+                controller.load_data(args.file)
+            except Exception:
+                logger.exception(f"Error while loading {args.file}")
+                raise SystemExit
```
Jc2k, line 71: Theres some stray whitespace on this line black is complaining about.

Jc2k (changes requested):
Thanks for this. I'll merge it as soon as the lints are happy.

As you can tell, use outside of homeassistant hasn't been a priority so I imagine there will be more fixes needed depending on what you are doing...

Jc2k (approved):

PR #390, opened 2025-02-28
Rework decrypt logic to reduce slicing

bdraco:
I doubt this will fix the issue https://github.com/home-assistant/core/issues/97888 but there is a lot less that can go wrong now since we only have one slice and one del


Jc2k (approved):

bdraco:
Thanks. Will do a release shortly.

PR #392, opened 2025-02-28
Modernize CI workflow

bdraco:
- codecov-action -> v5 to fix CI upload
- Use built-in poetry cache in setup-python v5
- Update checkout action to v4
- Add 3.13 to the CI


bdraco:
codecov is still not working because it needs an upload token but it will work once thats added

PR #403, opened 2025-03-29
Add pre-commit config and switch to ruff

bdraco:
- Add pre-commit config
- Reformat with ruff 
- flake8 remains at this time, a future PR can migrate to using ruff for linting, but there are too many issues to fix in this PR, and I wanted to leave this as basic ruff fixes + formatting only
- Strip trailing newlines
- Note that mypy and codespell are not turned on at this time because there are too many errors to fix.


bdraco:
retested on production to make sure there weren't any side effects.

All good

PR #433, opened 2025-04-22
Ensure connection is always shutdown when removing a pairing

### LINKED ISSUE #343: Out of order pairing steps when using HomeSpan (async race condition?)
HomeSpan is a library for ESP32 devices to enable them to speak HomeKit. After a huge amount of tests, it seems that when pairing these devices with Home Assistant via the HomeKit Device integration, pairing intermittently fails in a variety of ways. 

- Sometimes Home Assistant starts pairing and then complains that the device is already paired
- Sometimes it all works perfectly
- Sometimes Home Assistant fails on step `04`

I've included a massive amount of logs in my original HomeSpan bug report, but the description of what's happening from the devs makes it sound like there's some sort of race condition in aiohomekit which is causing the pairing steps to happen in a random order (nondeterministically).

The HomeSpan issue link is here: https://github.com/HomeSpan/HomeSpan/issues/680

but the important part I think is here:

> Thanks -- these details are helpful, though I'm not sure there is much I can do. As you can see from your first log output, the first thing HA is doing is asking for a pair-verify. That makes no sense - it first needs to set up pairing before it can open a secure paired channel. However, this won't cause a problem for HomeSpan as it just reports the error and waits for the next input. Based on your log file, HA does send the correct pair-setup request, but waits for over a minute before sending. I have no idea why HA would wait so long. As you can see, it correctly send the 3 steps for pairing in the correct order (M1, M3, and then M5) and the device properly pairs.
>
> However, 10 seconds later HA seems to open a new connection and tries to pair again. Since the device is already paired, HomeKit responds with an error in accordance with the HAP protocol. Next, HA tries to create a secure connection using the steps for a pair-verify. Since the device is paired, this works and the verification completes. I do not know why HA does not think the device is paired since it has gone as far to open a secure connection.
>
> As you noted, the second log file shows a correct pairing and verification as HA sent all the right steps in the correct order.
>
> In the third log file, HA seems to be trying to verify a secure connection before it has finished all the pairing steps. HAP does not specify what is supposed to happen if a device interrupts a pair-setup with a misplaced pair-verify. I coded HomeSpan to reset the pairing procedure whenever that occurs so things can start fresh, but it looks like HA tries to continue the pairing setup by sending an M3 (which is the second step of pairing) --- by that time HomeSpan has already reset the pairing sequence and is expected to see M1 (which is the first step).
>
> Based on the above logs, it almost seems as if HA is running two parallel processes to try and pair/verify, such that depending on the exact timing, if one process completes everything works, but if they interfere, the behavior is indeterminant.

Could this be a bug in aiohomekit? Or maybe the Home Assistant integration itself? HomeSpan devices always pair flawlessly with "real" HomeKit controllers (e.g. via the Home app) and claim to be fully compliant with the spec, so I'm not sure what's going on

bdraco:
- If anything fails, we should still explictly shutdown the connection.
- If the device gets rediscovered after shutdown via `_async_description_update`, don't reconnect

fixes #343

### DIFF
--- a/aiohomekit/controller/ip/pairing.py
+++ b/aiohomekit/controller/ip/pairing.py
@@ -501,7 +501,9 @@ class IpPairing(ZeroconfPairing):
             (TLV.kTLVType_Identifier, pairingId.encode("utf-8")),
         ]
 
+        logger.debug("Removing pairing with id %s", pairingId)
         data = dict(await self.connection.post_tlv("/pairings", request_tlv))
+        logger.debug("Remove pairing response: %s", data)
 
         if data.get(TLV.kTLVType_State, TLV.M2) != TLV.M2:
             raise InvalidError("Unexpected state after removing pairing request")
--- a/aiohomekit/controller/ip/pairing.py
+++ b/aiohomekit/controller/ip/pairing.py
@@ -503,7 +503,6 @@ class IpPairing(ZeroconfPairing):
 
         logger.debug("Removing pairing with id %s", pairingId)
         data = dict(await self.connection.post_tlv("/pairings", request_tlv))
-        logger.debug("Remove pairing response: %s", data)
 
         if data.get(TLV.kTLVType_State, TLV.M2) != TLV.M2:
             raise InvalidError("Unexpected state after removing pairing request")
--- a/aiohomekit/controller/ip/pairing.py
+++ b/aiohomekit/controller/ip/pairing.py
@@ -501,7 +501,6 @@ class IpPairing(ZeroconfPairing):
             (TLV.kTLVType_Identifier, pairingId.encode("utf-8")),
         ]
 
-        logger.debug("Removing pairing with id %s", pairingId)
         data = dict(await self.connection.post_tlv("/pairings", request_tlv))
 
         if data.get(TLV.kTLVType_State, TLV.M2) != TLV.M2:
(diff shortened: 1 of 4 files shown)


Jc2k (approved):

aiohomekit/controller/ip/pairing.py
```diff
@@ -501,7 +501,9 @@ async def remove_pairing(self, pairingId: str) -> bool:
             (TLV.kTLVType_Identifier, pairingId.encode("utf-8")),
         ]
 
+        logger.debug("Removing pairing with id %s", pairingId)
         data = dict(await self.connection.post_tlv("/pairings", request_tlv))
+        logger.debug("Remove pairing response: %s", data)
```
bdraco, line 506:
```suggestion
```

aiohomekit/controller/ip/pairing.py
```diff
@@ -501,6 +501,7 @@ async def remove_pairing(self, pairingId: str) -> bool:
             (TLV.kTLVType_Identifier, pairingId.encode("utf-8")),
         ]
 
+        logger.debug("Removing pairing with id %s", pairingId)
```
bdraco, line 504:
```suggestion
```

Jc2k (approved):

PR #498, opened 2025-09-13
Handle unknown HAP status codes gracefully

bdraco:
# Handle unknown HAP status codes gracefully

## Summary

Fixes crashes when accessories return undefined HAP status codes (e.g., Nanoleaf's `-6`). Previously, `to_status_code()` would raise `ValueError` for unknown codes, causing HomeKit Controller setup failures.

## Changes

- Added `HapStatusCode.UNKNOWN` enum value for undefined status codes
- Modified `to_status_code()` to return `UNKNOWN` instead of raising `ValueError`
- Added debug logging to track unknown status codes for future investigation

## Related Issues

- https://github.com/home-assistant/core/issues/152259
- Fixes Nanoleaf device setup failures with `ValueError: -6 is not a valid HapStatusCode`

### DIFF
--- a/aiohomekit/controller/ip/pairing.py
+++ b/aiohomekit/controller/ip/pairing.py
@@ -63,11 +63,12 @@ def format_characteristic_list(
     # Handle global error status first - set defaults for all requested characteristics
     if "status" in data and data["status"] != 0:
         # Device returned a global error status
-        try:
-            description = to_status_code(data["status"]).description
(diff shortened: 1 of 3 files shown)


bdraco:
I'm going to get these out so we don't miss the next HA patch release.

lroehrs:
Thanks for the patch @bdraco, is also fixing my issues! :)

PR #548, opened 2026-04-05
Update Ecobee definitions

sirwolfgang:
This pull request updates and expands the Ecobee-specific characteristic definitions in `characteristic_types.py`. The main changes include correcting UUID assignments for several thermostat characteristics, improving documentation for scheduled change time handling, and adding new characteristics for equipment status and SmartSensor data.

Updates to Ecobee characteristic UUIDs and documentation:

* Corrected UUID assignments for sleep/away heat and cool targets to match the Ecobee specification.
* Improved and clarified documentation for `VENDOR_ECOBEE_NEXT_SCHEDULED_CHANGE_TIME`, including details on value formats, meaning, and correct write behavior for timed holds.

New Ecobee characteristic definitions:

* Added new read-only characteristics for equipment running state, thermostat status code, and auxiliary heat activity.
* Introduced SmartSensor-specific characteristics for reporting seconds since last motion and occupancy activations.

### DIFF
--- a/aiohomekit/model/characteristics/characteristic_types.py
+++ b/aiohomekit/model/characteristics/characteristic_types.py
@@ -318,21 +318,25 @@ class CharacteristicsTypes:
     # r/w, float - home cool temperature between 18.3 and 33.3
     VENDOR_ECOBEE_HOME_TARGET_COOL = "7D381BAA-20F9-40E5-9BE9-AEB92D4BECEF"
     # r/w, float - sleep heat temperature between 7.2 and 26.1
-    VENDOR_ECOBEE_SLEEP_TARGET_HEAT = "73AAB542-892A-4439-879A-D2A883724B69"
+    VENDOR_ECOBEE_SLEEP_TARGET_HEAT = "05B97374-6DC0-439B-A0FA-CA33F612D425"
     # r/w, float - sleep cool temperature between 18.3 and 33.3
-    VENDOR_ECOBEE_SLEEP_TARGET_COOL = "5DA985F0-898A-4850-B987-B76C6C78D670"
+    VENDOR_ECOBEE_SLEEP_TARGET_COOL = "A251F6E7-AC46-4190-9C5D-3D06277BDF9F"
     # r/w, float - away heat temp between 7.2 and 26.1
-    VENDOR_ECOBEE_AWAY_TARGET_HEAT = "05B97374-6DC0-439B-A0FA-CA33F612D425"
+    VENDOR_ECOBEE_AWAY_TARGET_HEAT = "73AAB542-892A-4439-879A-D2A883724B69"
     # r/w, float - away cool temp between 18.3 and 33.3
-    VENDOR_ECOBEE_AWAY_TARGET_COOL = "A251F6E7-AC46-4190-9C5D-3D06277BDF9F"
+    VENDOR_ECOBEE_AWAY_TARGET_COOL = "5DA985F0-898A-4850-B987-B76C6C78D670"
     # w/o, uint8 - set hold schedule mode - home(0)/sleep(1)/away(2)
     VENDOR_ECOBEE_SET_HOLD_SCHEDULE = "1B300BC2-CFFC-47FF-89F9-BD6CCF5F2853"
-    # r/w, string - Next scheduled event time, this time seems to be a local timestamp formatted similar to "2024-10-19T20:35:05-04:00Q"
-    # This would be when the currently scheduled activity ends such as:
-    # 1. If there is a temporary hold in place, will show when that hold ends (the default hold is until next scheduled activity, so usually behaves like 2)
-    # 2. If there are no holds in place will show the next scheduled climate mode time
-    # 3. If there is an indefinite hold, will show a generally far-future date like "2035-12-29T12:00:00-04:00Q"
-    # NOTE: When setting this value, it seems to be interpreted as a local time so recommend using formatting like "2024-01-01T12:00:00" without an offset or you may get weird results
+    # r/w, string - Timed hold end time or next scheduled event time.
+    # Ecobee reports values with a trailing 'T' suffix, e.g. "2024-10-19T20:35:05-04:00T".
+    # Values:
+    # 1. Timed hold active: ISO-8601 local time when hold ends, trailing T suffix
+    # 2. No hold (following schedule): next scheduled climate change time, trailing T suffix
+    # 3. Indefinite hold: far-future sentinel, typically "2035-01-03T00:00:00+00:00T"
+    # Writing: set a timed hold end time. Must be written BEFORE SET_HOLD_SCHEDULE in a
+    # separate HAP request; combined writes cause ecobee to ignore TIMESTAMP and default
+    # to permanent hold. Write format: ISO-8601 with local offset + trailing T suffix,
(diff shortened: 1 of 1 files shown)


bdraco:
Please explain the history and how the the existing ones were incorrect

sirwolfgang:
@bdraco I don't understand how I can explain the history, I didn't create the first version. The diff shows how they were swapped.

bdraco:
> @bdraco I don't understand how I can explain the history, I didn't create the first version. The diff shows how they were swapped.

Unfortunately without understanding why they were set to the UUIDs currently in use, I can't merge this PR because I don't know if its going to break older/different models.

sirwolfgang:
@Jc2k When you bulk introduced these IDs 4 years ago, do you remember where you got them? And certain that they were all set correct? _i'd wouldn't be able to answer this if it was me, 4 years is a long time_

@bdraco I understand what your trying to maintain with that position, it is true that my device is newer than 4 years old; but in the realm of known knowns, known unknowns, and unknown unknowns, these are the facts

1. It is unknown if these UUIDs were labeled correctly
2. It is known that these UUIDs are incorrect today for Ecobee units connected to the internet today, running the current version of firmware.

Therefore the risk you are trying to protect is a cohort of past units (less than present and future units) that is not connected to the internet from being affected by a software patch to a system on the same network, that is connected to the internet. That is to say, if the both the old and new versions are correct in their place in time, updating this to reflect the current version would not break those on the old version because those networks would still be on the old versions together; All of which assumes that the old version is correct which is a known unknown.

coffee-the-dev:
Came across this while digging into better Home Assistant support for a HomeKit-connected ecobee, especially around trying to expose more of the raw datapoints that already show up in HomeKit diagnostics.

This seems related to https://github.com/home-assistant/core/pull/167487 which looks like it depended on additional/corrected aiohomekit characteristic definitions.

Is there any path forward for this PR? I can provide current ecobee HomeKit diagnostics with the UUIDs, model/firmware, and observed values for my own ecobee if that helps validate anything here.

bharvey88:
@bdraco re the history you asked for: the current labels trace to a single unverified 2020 comment, and every independent source from 2017 through 2026 matches this PR's mapping, so there's no sign older models ever used the current labels. Details below.

**Where the current labels came from.** These UUIDs were first posted unlabeled in [home-assistant/core#30258 (Jan 2020)](https://github.com/home-assistant/core/issues/30258#issuecomment-573108986), with the pairing an open question ("I wonder if the 3 pairs of temperature characteristic are the upper and lower thresholds for away, home and sleep mode?"). [A follow-up comment in that thread (Jun 2020)](https://github.com/home-assistant/core/issues/30258#issuecomment-647160757), from an ecobee3 lite, annotated them as sleep = `73AAB542`/`5DA985F0` and away = `05B97374`/`A251F6E7` without saying how that was determined. #49 (Dec 2021) then added the constants with that comment's text word for word ("sleep heat temperature between 7.2 and 26.1", "away heat temp between 7.2 and 26.1"). As far as I can find, that one comment is the only source for the current mapping, and everything else that agrees with it ([HA's const.py](https://github.com/home-assistant/core/blob/dev/homeassistant/components/homekit_controller/const.py), [dk307/HSPI_HomeKitControllerLite](https://github.com/dk307/HSPI_HomeKitControllerLite), a couple of ISY plugin repos) postdates #49 and takes its names from aiohomekit.

**Independent evidence for the swapped mapping.** forty2/hap-client-tool committed [src/metadata/ecobee3.json](https://github.com/forty2/hap-client-tool/blob/master/src/metadata/ecobee3.json) on 2017-05-15, four and a half years before #49 and outside this lineage. It names `73AAB542` "Away Comfort Heat Level", `5DA985F0` "Away Comfort Cooling Level", `05B97374` "Sleep Comfort Heat Level", `A251F6E7` "Sleep Comfort Cooling Level", and agrees with both sides on the Home pair. That's this PR's mapping, in the first-gen firmware era.

**Values across firmware eras.** No known dump shows ecobee sending `description` strings for its vendor characteristics (it sends them for Apple-standard ones only), so setpoint values are the next best signal. In every real dump I could find, the `73AAB542`/`5DA985F0` pair is the deeper setback (values converted from the Celsius in the dumps):

| Device, firmware, source | 73AAB542 / 5DA985F0 | 05B97374 / A251F6E7 |
|---|---|---|
| ecobee3, 4.2.394, [2019 HA fixture](https://github.com/home-assistant/core/blob/dev/tests/components/homekit_controller/fixtures/ecobee3.json) (same dump as this repo's `tests/fixtures/ecobee3.json`) | 64 / 82 F | 66 / 80 F |
| ECB501, 4.7.340214, [HA fixture](https://github.com/home-assistant/core/blob/dev/tests/components/homekit_controller/fixtures/ecobee_501.json) | 64 / 82 F | 65 / 82 F |
| ecobee3 lite, 4.7.430164, [diagnostics in home-assistant/core#85715](https://github.com/home-assistant/core/issues/85715) (2023) | 50 / 85 F | 65 / 76 F |
| ecobee3 lite, 4.8.70226, [2025 HA fixture](https://github.com/home-assistant/core/blob/dev/tests/components/homekit_controller/fixtures/ecobee3_lite.json) | 64 / 78 F | 68 / 76 F |

The 50/85 F row is hard to read any other way: 85 F is ecobee's [default Away cool threshold](https://support.ecobee.com/hc/en-us/articles/115007463208-What-are-comfort-settings-and-how-does-scheduling-work-on-my-ecobee-), and nobody sleeps with heat set to 50 F. So there's no sign of a firmware-side meaning change between 2017 and 2026; the labels look transposed since introduction. A plausible mechanism: the setpoints appear in accessory order home, away, sleep (iids 34 to 39), while the hold enum is home(0)/sleep(1)/away(2), and labeling the pairs in enum order produces exactly this swap.

**The five added characteristics** are not new-firmware additions. All of them are already in the 2019 dump with the same formats, perms and ranges (`4A6AE4F6` uint8 0-4, `DB7BF261` uint8 0-3, `41935E3E` bool, both last-activation ints). The two last-activation ones only ever appear inside motion/occupancy sensor services (a 2026 dump also carries `"unit": "seconds"`, range -1 to 86400), which corroborates them structurally. The three thermostat-status ones exist on every model dumped, but their claimed semantics rest on live observation of one unit, so the diagnostics offered above would be a useful cross-check.

bharvey88:
Opened a comment-only PR against the source branch with the two doc fixes discussed: [sirwolfgang/aiohomekit#1](https://github.com/sirwolfgang/aiohomekit/pull/1). Merging that into `patch/ecobee` will update this PR.

bharvey88:
@sirwolfgang I spoke with bdraco yesterday and he suggested I submit the PR to your repo above^ Hopefully this will be able to get merged once you merge my PR.

sirwolfgang:
Merged

bluetoothbot (approved):
> [!TIP]

bharvey88:
Second comment-only PR on the source branch addressing the two review suggestions (provenance notes for the status enums and the write-format guidance): [sirwolfgang/aiohomekit#2](https://github.com/sirwolfgang/aiohomekit/pull/2).

sirwolfgang:
Merged

bdraco (approved):

bdraco:
Thanks @sirwolfgang @bharvey88

PR #549, opened 2026-04-19
Replace bleak_scanner_instance with bleak_scanner_factory

bdraco:
bleak removed register_detection_callback, so the detection callback must now be passed via the BleakScanner constructor. There is no way to add it after construction, so accepting an already built scanner no longer works.

Switch the Controller API to accept a factory; the factory takes the detection callback and returns the scanner. Callers like Home Assistant can now build the wrapper at the right moment, e.g. partial(bluetooth.async_get_scanner, hass).

This is a breaking change to the Controller signature, but it is forced by the upstream bleak removal; there is no compatible alternative.

### DIFF
--- a/aiohomekit/controller/ble/controller.py
+++ b/aiohomekit/controller/ble/controller.py
@@ -2,11 +2,11 @@ from __future__ import annotations
 
 import asyncio
 import logging
-from collections.abc import AsyncIterable
+from collections.abc import AsyncIterable, Callable
 
 from bleak import BleakScanner
 from bleak.backends.device import BLEDevice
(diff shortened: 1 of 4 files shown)


elupus:
Why are this done? Why not just construct a standard BleakScanner() the standard way. We already patch that to be the home assistant type? (Ps, this i partially why i liked passing a backend instead :))

bdraco:
We had an ordering problem before where it wasn't installed in time.  Sadly I'm leaving this afternoon to be traveling until after beta ships and this is currently broken.  I'm out of time so I'm going to fallback to doing a revert of habluetooth instead

PR #552, opened 2026-04-21
Drop bleak_scanner_instance; construct BleakScanner internally

### LINKED ISSUE #556: register_detection_callback deprecated in habluetooth — needs design fix
## Problem

`habluetooth` 6.x deprecated `HaBleakScannerWrapper.register_detection_callback()` with this warning:

```
HaBleakScannerWrapper.register_detection_callback() is deprecated and will be removed in a future release;
bleak already removed this method from BleakScanner. Pass detection_callback to the HaBleakScannerWrapper constructor instead.
```

This warning fires on every Home Assistant startup because `aiohomekit/controller/ble/controller.py` calls `self._scanner.register_detection_callback(self._device_detected)` in `async_start()`.

## Root cause

The warning is straightforward to fix when `aiohomekit` creates its own scanner:
```python
# current
self._scanner = BleakScanner()
self._scanner.register_detection_callback(self._device_detected)

# fix
self._scanner = BleakScanner(detection_callback=self._device_detected)
```

However, when an external scanner is passed in (as Home Assistant does via `bluetooth.async_get_scanner(hass)`), the scanner is already constructed before `async_start()` is called — so the callback cannot be passed at construction time. The only current option is `register_detection_callback()`, which is deprecated.

## Why HA can't fix it unilaterally

Home Assistant passes a `HaBleakScannerWrapper` to `Controller` so that all integrations share a single centralized BLE scanner. HA doesn't know the detection callback at the time the scanner is created (the callback is internal to `BleController`). So HA cannot pre-configure the scanner with the callback.

## Possible solutions

1. **Scanner factory parameter**: Add a `bleak_scanner_factory: Callable[[AdvertisementDataCallback], BleakScanner] | None` parameter to `Controller` (or `BleController`). HA would pass a factory instead of an instance. `aiohomekit` calls the factory with the callback at `async_start()` time.

2. **Expose the detection callback**: Add a public `ble_detection_callback` property to `Controller`/`BleController` so callers can read it and construct the scanner themselves before passing it in.

3. **Coordinate with habluetooth**: Ask `habluetooth` to add a non-deprecated `update_detection_callback()` method for the case where the scanner is shared and must be updated post-construction.

Option 1 seems cleanest and is fully backwards-compatible (old code passing `bleak_scanner_instance` still works; new code can pass a factory instead).

## Affected versions

- `aiohomekit` 3.2.20
- `habluetooth` 6.1.0
- Home Assistant 2026.5.2 (`components/homekit_controller/utils.py:82`)

bdraco:
bleak removed `register_detection_callback`, so the detection callback must now be passed via the `BleakScanner` constructor. There is no way to add it after construction, so accepting an already-built scanner no longer works.

Drop the `bleak_scanner_instance` parameter from `Controller` and `BleController` entirely and always construct `BleakScanner(detection_callback=self._device_detected)` internally in `BleController.async_start`. Also drop the now-unused `register_detection_callback(None)` call from `async_stop`.

This is safe for Home Assistant because `habluetooth` already overwrites `BleakScanner` by the time we construct it, so the internally-created scanner is the HA-managed one.

This is a breaking change to the `Controller` signature, but it is forced by the upstream bleak removal; there is no compatible alternative.

fixes #556


bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Dropping the scanner-instance escape hatch narrows when BLE gets registered

bdraco:
The BLE gating is safe for HA; habluetooth imports bleak long before aiohomekit is loaded, so bleak is always in sys.modules when aiohomekit.const is imported, and AIOHOMEKIT_TRANSPORT_BLE remains as an explicit opt in for consumers with different import ordering. The debug level logging on scanner init and start failure predates this change and is intentional; systems without BLE hardware hit these paths on every start and a warning would be noise. The lock file is now regenerated with Poetry 2.4.1 so the only lock delta is the bleak bump.

@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
Restored the idempotency guard; a second async_start now returns early instead of replacing the running scanner, with a test covering it. Keeping the debug level on scanner init and start failures; BLE registration already requires opting in via an early bleak import or AIOHOMEKIT_TRANSPORT_BLE, and HA hosts without a bluetooth adapter go through this path on every start, so a warning would be recurring noise for an expected condition.

@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

PR #579, opened 2026-07-22
Try the next advertised address when pair-verify reaches the wrong accessory

### LINKED ISSUE #571: IncorrectPairingIdError at Pair-Verify step 3 should try next  address from mDNS addresses list instead of retrying same IP
## Summary

When a HAP accessory (category: Bridge) advertises multiple IP 
addresses in its mDNS record and the first address returns 
`IncorrectPairingIdError` at Pair-Verify step 3, aiohomekit 
retries indefinitely on the same IP instead of advancing to 
the next address in the list.

## Device

BTicino Smarther 2 (model XW8002) — HAP Bridge (category=2)
Firmware 40.0.0

## Root cause

The Smarther 2 firmware bug: each thermostat advertises ALL 
thermostats' IPs in its mDNS `addresses` list, with rotating 
order on each re-announcement.

Example mDNS record for device `d2:5a:0e:93:a7:7c` (real IP: .10):
addresses: ['192.168.2.13', '192.168.2.10', '192.168.2.14', '192.168.2.11']
First address `.13` belongs to a different physical device 
(`70:25:1d:30:8e:38`), which correctly returns its own 
PairingID at step 3 — causing `IncorrectPairingIdError` 
for the requesting controller.

## Observed behavior
Attempting connection to ['192.168.2.13', '192.168.2.10', ...]:5001

→ pair-verify on 192.168.2.13 → step 3 FAIL (IncorrectPairingIdError)

→ Retrying in 0s
Attempting connection to ['192.168.2.13', '192.168.2.10', ...]:5001

→ pair-verify on 192.168.2.13 → step 3 FAIL (IncorrectPairingIdError)

→ Retrying in 1s

[infinite loop on same IP]

The correct IP (`.10`) is at index 1 of the list, but aiohomekit 
never advances past index 0.

## Expected behavior (per Apple HAP behavior)

When `IncorrectPairingIdError` is raised at step 3, the 
controller should try the next address in the `addresses` 
list rather than retrying the same IP. Apple's iOS/HomePod 
implementation handles this gracefully — the same devices 
work reliably with Apple Home despite the firmware bug.

## Suggested fix

In `aiohomekit/controller/ip/connection.py`, in the reconnect 
loop, when `IncorrectPairingIdError` is raised at step 3:
- advance to the next address in the addresses list
- only raise/retry from start after exhausting all addresses

## Current workaround

We deployed a Python sidecar (using `zeroconf`) that monitors 
`_hap._tcp.local.`, intercepts multi-address records, and 
re-registers corrected `ServiceInfo` objects containing only 
the single correct IP per device. This prevents aiohomekit 
from ever seeing the wrong IP as first in the list.

## Environment

- Home Assistant 2026.6.2
- aiohomekit (bundled)
- Python 3.14
- 5x BTicino Smarther 2 XW8002

bdraco:
## Summary

Some buggy bridges like the BTicino Smarther 2 advertise other devices' IPs in their mDNS record, with the order rotating on each announcement. When the first address answers, TCP connects fine but pair-verify step 3 returns another accessory's pairing id, and the reconnect loop kept retrying the same address forever; the correct address later in the list was never reached.

## Changes

When pair-verify raises IncorrectPairingIdError the connected address is remembered as belonging to a different accessory and is skipped on the next attempt; the reconnect loop retries right away while unmarked addresses remain, otherwise the normal backoff applies. If every address has been marked, or the advertised address set changes, the exclusions are reset so the accessory can never become permanently unreachable.

## Testing

New tests cover host filtering, marking the failed address, recovery after the exclusions reset, the immediate retry, and clearing on an address change; full suite passes locally.

fixes #571


bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

PR #580, opened 2026-07-22
Cover the normalize host fallback for hostnames

bdraco:
## Summary

Followup to #579; codecov flagged the ValueError fallback in _normalize_host as uncovered, the path taken when an advertised host is not an IP address.

## Changes

Add a test that excludes a hostname through _get_connect_hosts so the fallback is exercised, and add the missing return annotations on the tests from #579.

## Testing

Full suite passes locally and the two flagged lines are now covered.


bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

PR #581, opened 2026-07-22
Clear the BLE client when disconnect fails with a dead dbus socket

bdraco:
## Summary

When the D-Bus socket dies, disconnect raises EOFError and _close_while_locked only caught BleakError, so self.client was never cleared; is_connected then reported a stale True forever and _ensure_connected never reconnected, which is the permanent BLE stall in #534. The root cause is also being fixed upstream in hbldh/bleak#2012, and Bluetooth-Devices/bleak-retry-connector#309 makes transient socket errors retry; this change makes aiohomekit recover on any released bleak version.

## Changes

_close_while_locked now catches the broad retry exception set, which includes EOFError and BrokenPipeError, so the client is always cleared and connection state reset after a failed disconnect.

## Testing

New tests cover a client whose disconnect raises EOFError and one raising BleakError, asserting the client is cleared in both cases; full suite passes.

fixes #534

### DIFF
--- a/aiohomekit/controller/ble/pairing.py
+++ b/aiohomekit/controller/ble/pairing.py
@@ -903,7 +903,10 @@ class BlePairing(AbstractPairing):
             return
         try:
             await self.client.disconnect()
-        except BleakError:
+        except BLEAK_EXCEPTIONS:
+            # A dead D-Bus socket raises EOFError, so catch the broad retry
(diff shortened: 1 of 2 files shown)


bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

PR #582, opened 2026-07-23
Skip BLE characteristics that return a success PDU without a value

bdraco:
## Summary

Some accessories, such as the UltraLoq Bolt Fingerprint, return a success CHAR_READ PDU whose TLV has no value parameter; the hard subscript in _get_characteristics_while_connected raised KeyError: 1 and failed the whole config entry setup in Home Assistant, see home-assistant/core#168578. Apple Home tolerates the same device.

## Changes

The read loop now looks the value up with get and skips the characteristic with a debug log when it is missing, matching the existing PDUStatusError and struct.error skip paths on either side of it; _populate_char_values already tolerates a characteristic without a value. The pairing management paths that hard index the same TLV key are intentionally untouched, they should fail loudly.

## Testing

New test drives the read loop with one valid response and one success response without a value TLV, asserting the good characteristic decodes and the bad one is skipped; the test reproduces the KeyError on the previous code, full suite passes.

### DIFF
--- a/aiohomekit/controller/ble/pairing.py
+++ b/aiohomekit/controller/ble/pairing.py
@@ -75,6 +75,7 @@ from .client import (
     raise_for_pdu_status,
 )
 from .connection import establish_connection
+from .const import AdditionalParameterTypes
 from .key import BroadcastDecryptionKey, DecryptionKey, EncryptionKey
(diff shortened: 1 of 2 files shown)


bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

PR #583, opened 2026-07-23
Skip linked service references to iids that do not exist

bdraco:
## Summary

The Yeelight Pro Gateway exposes ControlPanel accessories whose service label links reference service iids that do not exist in the accessory, which raised KeyError while parsing the accessories payload and prevented the whole bridge from pairing in Home Assistant, see home-assistant/core#175416. The gateway then reports itself as paired so users have to reset it to retry.

## Changes

The linked service loop in Accessory.create_from_dict now resolves each linked iid with iid_or_none and skips references that do not exist, extending the existing Schlage Encode Plus guard for iid 0 on the same lines; serialization only emits resolved services so the dropped references stay gone. The parent service lookup is hoisted out of the inner loop while touching it.

## Testing

New test parses an accessory with a linked list containing iid 0, a dangling iid, and a real one, asserting only the real service is linked and that serialization round trips without the dangling reference; the test reproduces the KeyError on the previous code, full suite passes.

### DIFF
--- a/aiohomekit/model/__init__.py
+++ b/aiohomekit/model/__init__.py
@@ -304,13 +304,18 @@ class Accessory:
                     char.set_value(char_data["value"])
 
         for service_data in data["services"]:
-            for linked_service in service_data.get("linked", []):
+            if not (linked_iids := service_data.get("linked")):
+                continue
+            service = accessory.services.iid(service_data["iid"])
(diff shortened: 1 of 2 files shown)


bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

